PR #15, opened 2020-08-17
Added the initial draft of the service, probe and targets proto files.

### LINKED ISSUE #4: Write Service Config File.
# Objective
Write the configuration file for Hermes as a service.

### LINKED ISSUE #9: Create Hermes API config file

evanSpendlove:
This is linked to two issues:
- Closes #4 - Write Service Config file .
- Closes #9 - Write Hermes API config file.

### DIFF
--- a/config/proto/probe.proto
+++ b/config/proto/probe.proto
@@ -12,7 +12,7 @@ import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
 
 option go_package = "github.com/googleinterns/step224-2020/config/proto";
 
-// This defines the top-level probe interface for Hermes probes.
+// HermesProbeDef defines the top-level probe interface for Hermes probes.
 // All Hermes probes should extend this probe config.
 message HermesProbeDef {
   string probe_name= 1;
--- a/config/proto/probe.proto
+++ b/config/proto/probe.proto
@@ -16,11 +16,11 @@ message HermesProbeDef {
     CEPH = 2;
   }
 
-  // Type of Probe
+  // Probe types currently supported
   enum ProbeType {
-    WRITE = 1;
-    DELETE = 2;
-    CHECK = 3;
+    GET = 1; // GET probe is for getting an object (file) from a storage system.
+    PUT = 2; // PUT probe is for putting an object (file) into a storage system.  
+    DELETE = 3; // DELETE probe is for deleting an object (file) from a storage system.
   }
 
   TargetSystem target_system = 3;
@@ -32,8 +32,8 @@ message HermesProbeDef {
   optional int32 timeout_sec = 6 [default = 60]; // Timeout in seconds
 
   // If specified, latency is stored as a distribution
-  optional metrics.Dist probe_latency_distribution = 7;
-  optional metrics.Dist api_call_latency_distribution = 8;
+  optional metrics.Dist probe_latency_distribution = 7; // This measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
+  optional metrics.Dist api_call_latency_distribution = 8; // This measures the latency of the API calls for the relevant storage system targeted.
 
   // Valid values: "ns", "us" (or "µs"), "ms", "s", "m", "h".
   optional string latency_unit = 9 [default = "us"];
--- a/config/proto/probe.proto
+++ b/config/proto/probe.proto
@@ -1,6 +1,8 @@
 /*
   This proto defines the initial probe interface for Hermes probes.
   It is used in a variety of other files as it is the top-level probe interface.
+  
+  Author: Evan Spendlove (@evanSpendlove)
 */
 syntax = "proto3";
 
@@ -41,8 +43,8 @@ message HermesProbeDef {
   int32 timeout_sec = 6 [default = 60];  // Timeout in seconds
 
   // If specified, latency is stored as a distribution
-  metrics.Dist probe_latency_distribution = 7;  // This measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
-  metrics.Dist api_call_latency_distribution = 8;  // This measures the latency of the API calls for the relevant storage system targeted.
+  metrics.Dist probe_latency_distribution = 7;  // Measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
+  metrics.Dist api_call_latency_distribution = 8;  // Measures the latency of the API calls for the relevant storage system targeted.
 
   // Valid values: "ns", "us" (or "µs"), "ms", "s", "m", "h".
   string latency_unit = 9 [default = "us"];
--- a/config/proto/probe.proto
+++ b/config/proto/probe.proto
@@ -3,6 +3,7 @@ syntax = "proto3";
 package = hermes;
 
 import "github.com/google/cloudprober/metrics/proto/dist.proto";
+import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
 
 option go_package = "github.com/googleinterns/step224-2020/config/proto"
 
--- a/config/proto/probe.proto
+++ b/config/proto/probe.proto
@@ -1,6 +1,6 @@
 syntax = "proto3";
 
-package = hermes;
+package hermes;
 
 import "github.com/google/cloudprober/metrics/proto/dist.proto";
 import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
@@ -49,4 +49,3 @@ message AdditionalLabel {
   required string key = 1;
   required string value = 2;
 }
-
--- a/config/proto/probe.proto
+++ b/config/proto/probe.proto
@@ -5,7 +5,7 @@ package = hermes;
 import "github.com/google/cloudprober/metrics/proto/dist.proto";
 import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
 
-option go_package = "github.com/googleinterns/step224-2020/config/proto"
+option go_package = "github.com/googleinterns/step224-2020/config/proto";
 
 message HermesProbeDef {
   required string probe_name= 1;
--- a/config/proto/probe.proto
+++ b/config/proto/probe.proto
@@ -6,16 +6,20 @@ syntax = "proto3";
 
 package hermes;
 
+import "github.com/google/cloudprober/probes/proto/config.proto";
 import "github.com/google/cloudprober/metrics/proto/dist.proto";
 import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
 
 option go_package = "github.com/googleinterns/step224-2020/config/proto";
-  // This defines the top-level probe interface for Hermes probes.
-  // All Hermes probes should extend this probe config.
+
+// This defines the top-level probe interface for Hermes probes.
+// All Hermes probes should extend this probe config.
 message HermesProbeDef {
   string probe_name= 1;
-  targets.TargetsDef targets = 2;
+  TargetsDef targets = 2;
 
+  // TargetSystem is expected to be a storage system.
+  // It will stay constant until support for new storage systems is added.
   enum TargetSystem {
     GCS = 1;
     CEPH = 2;
@@ -23,35 +27,28 @@ message HermesProbeDef {
 
   // Probe types currently supported
   enum ProbeType {
-    GET = 1; // GET probe is for getting an object (file) from a storage system.
-    PUT = 2; // PUT probe is for putting an object (file) into a storage system.  
-    DELETE = 3; // DELETE probe is for deleting an object (file) from a storage system.
+    GET = 1;  // GET probe is for getting an object (file) from a storage system.
+    PUT = 2;  // PUT probe is for putting an object (file) into a storage system.  
+    DELETE = 3;  // DELETE probe is for deleting an object (file) from a storage system.
   }
 
   TargetSystem target_system = 3;
   ProbeType probe_type = 4;
 
-  int32 interval_sec = 5 [default = 3600]; // Probing interval in seconds
+  int32 interval_sec = 5 [default = 3600];  // Probing interval in seconds
 
   // Probes will not retry, they will fail if they timeout.
-  int32 timeout_sec = 6 [default = 60]; // Timeout in seconds
+  int32 timeout_sec = 6 [default = 60];  // Timeout in seconds
 
   // If specified, latency is stored as a distribution
-  metrics.Dist probe_latency_distribution = 7; // This measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
-  metrics.Dist api_call_latency_distribution = 8; // This measures the latency of the API calls for the relevant storage system targeted.
+  metrics.Dist probe_latency_distribution = 7;  // This measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
+  metrics.Dist api_call_latency_distribution = 8;  // This measures the latency of the API calls for the relevant storage system targeted.
 
   // Valid values: "ns", "us" (or "µs"), "ms", "s", "m", "h".
   string latency_unit = 9 [default = "us"];
 
   // Additional labels for the probe latency and API call latency distribution metrics.
   // Add as key-value pairs
-  repeated AdditionalLabel probe_latency_additional_label = 10;
-  repeated AdditionalLabel api_call_latency_additional_label = 11;
-}
-
-// Simple Key-Value pair for additional metric labels
-// Uses key, value instead of Map to allow for `repeated` keyword.
-message AdditionalLabel {
-  string key = 1;
-  string value = 2;
+  repeated cloudprober.probes.AdditionalLabel probe_latency_additional_label = 10;
+  repeated cloudprober.probes.AdditionalLabel api_call_latency_additional_label = 11;
 }
--- a/config/proto/probe.proto
+++ b/config/proto/probe.proto
@@ -1,0 +1,51 @@
+syntax = "proto3";
+
+package = hermes;
+
+import "github.com/google/cloudprober/metrics/proto/dist.proto";
+
+option go_package = "github.com/googleinterns/step224-2020/config/proto"
+
+message HermesProbeDef {
+  required string probe_name= 1;
+  required targets.TargetsDef targets = 2;
+
+  enum TargetSystem {
+    GCS = 1;
+    CEPH = 2;
+  }
+
+  // Type of Probe
+  enum ProbeType {
+    WRITE = 1;
+    DELETE = 2;
+    CHECK = 3;
+  }
+
+  TargetSystem target_system = 3;
+  ProbeType probe_type = 4;
+
+  optional int32 interval_sec = 5 [default = 3600]; // Probing interval in seconds
+
+  // Probes will not retry, they will fail if they timeout.
+  optional int32 timeout_sec = 6 [default = 60]; // Timeout in seconds
+
+  // If specified, latency is stored as a distribution
+  optional metrics.Dist probe_latency_distribution = 7;
+  optional metrics.Dist api_call_latency_distribution = 8;
+
+  // Valid values: "ns", "us" (or "µs"), "ms", "s", "m", "h".
+  optional string latency_unit = 9 [default = "us"];
+
+  // Additional labels for the probe latency and API call latency distribution metrics.
+  // Add as key-value pairs
+  repeated AdditionalLabel probe_latency_additional_label = 10;
+  repeated AdditionalLabel api_call_latency_additional_label = 11;
+}
+
+// Simple Key-Value pair for additional metric labels
+message AdditionalLabel {
+  required string key = 1;
+  required string value = 2;
+}
+
--- a/config/proto/probe.proto
+++ b/config/proto/probe.proto
@@ -13,8 +13,8 @@ option go_package = "github.com/googleinterns/step224-2020/config/proto";
   // This defines the top-level probe interface for Hermes probes.
   // All Hermes probes should extend this probe config.
 message HermesProbeDef {
-  required string probe_name= 1;
-  required targets.TargetsDef targets = 2;
+  string probe_name= 1;
+  targets.TargetsDef targets = 2;
 
   enum TargetSystem {
     GCS = 1;
@@ -31,17 +31,17 @@ message HermesProbeDef {
   TargetSystem target_system = 3;
   ProbeType probe_type = 4;
 
-  optional int32 interval_sec = 5 [default = 3600]; // Probing interval in seconds
+  int32 interval_sec = 5 [default = 3600]; // Probing interval in seconds
 
   // Probes will not retry, they will fail if they timeout.
-  optional int32 timeout_sec = 6 [default = 60]; // Timeout in seconds
+  int32 timeout_sec = 6 [default = 60]; // Timeout in seconds
 
   // If specified, latency is stored as a distribution
-  optional metrics.Dist probe_latency_distribution = 7; // This measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
-  optional metrics.Dist api_call_latency_distribution = 8; // This measures the latency of the API calls for the relevant storage system targeted.
+  metrics.Dist probe_latency_distribution = 7; // This measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
+  metrics.Dist api_call_latency_distribution = 8; // This measures the latency of the API calls for the relevant storage system targeted.
 
   // Valid values: "ns", "us" (or "µs"), "ms", "s", "m", "h".
-  optional string latency_unit = 9 [default = "us"];
+  string latency_unit = 9 [default = "us"];
 
   // Additional labels for the probe latency and API call latency distribution metrics.
   // Add as key-value pairs
@@ -52,6 +52,6 @@ message HermesProbeDef {
 // Simple Key-Value pair for additional metric labels
 // Uses key, value instead of Map to allow for `repeated` keyword.
 message AdditionalLabel {
-  required string key = 1;
-  required string value = 2;
+  string key = 1;
+  string value = 2;
 }
--- a/config/proto/probe.proto
+++ b/config/proto/probe.proto
@@ -1,3 +1,7 @@
+/*
+  This proto defines the initial probe interface for Hermes probes.
+  It is used in a variety of other files as it is the top-level probe interface.
+*/
 syntax = "proto3";
 
 package hermes;
@@ -6,7 +10,8 @@ import "github.com/google/cloudprober/metrics/proto/dist.proto";
 import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
 
 option go_package = "github.com/googleinterns/step224-2020/config/proto";
-
+  // This defines the top-level probe interface for Hermes probes.
+  // All Hermes probes should extend this probe config.
 message HermesProbeDef {
   required string probe_name= 1;
   required targets.TargetsDef targets = 2;
@@ -45,6 +50,7 @@ message HermesProbeDef {
 }
 
 // Simple Key-Value pair for additional metric labels
+// Uses key, value instead of Map to allow for `repeated` keyword.
 message AdditionalLabel {
   required string key = 1;
   required string value = 2;
--- a/config/proto/service.proto
+++ b/config/proto/service.proto
@@ -10,8 +10,9 @@ import "github.com/googleinterns/step224-2020/config/proto/probe.proto";
 import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
 
 option go_package = "github.com/googleinterns/step224-2020/config/proto";
+
 // This defines the (external) service-level config for Hermes.
-// It identifies what gRPCs Hermes will support.
+// Hermes identifies what gRPCs Hermes will support.
 service Hermes {
 
   // Start monitoring a new storage system.
@@ -33,10 +34,9 @@ message HermesProbeResponse {}
 
 // A target must be specified so that Hermes knows which target to stop monitoring.
 message StopMonitoringSystemRequest {
-  repeated targets.TargetsDef target = 1;
+  repeated TargetsDef target = 1;
 }
-// This is an empty response message for the probe response.
-// An exit status code may be added to probe responses, and included here.
+// TODO(#29) Add exit status for probe responses.
 message StopMonitoringSystemResponse {}
 
 // Empty message for listing monitoring systems as no arguments needed.
@@ -44,5 +44,5 @@ message ListMonitoredSystemsRequest {}
 
 // This message holds the list of targets returned by ListMonitoredStorageSystems()
 message ListMonitoredSystemsResponse {
-  repeated targets.TargetsDef target = 1;
+  repeated TargetsDef target = 1;
 }
--- a/config/proto/service.proto
+++ b/config/proto/service.proto
@@ -12,6 +12,7 @@ service Hermes {
   // Start monitoring a new storage system.
   rpc StartMonitoringStorageSystem(HermesProbeRequest) returns (HermesProbeResponse) {}
 
+  // Stop monitoring a storage system that is currently being monitored.
   rpc StopMonitoringStorageSystem(StopMonitoringSystemRequest) returns (StopMonitoringSystemResponse) {}
 
   // Lists the storage systems being monitored at the moment.
--- a/config/proto/service.proto
+++ b/config/proto/service.proto
@@ -12,7 +12,7 @@ import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
 option go_package = "github.com/googleinterns/step224-2020/config/proto";
 
 // This defines the (external) service-level config for Hermes.
-// Hermes identifies what gRPCs Hermes will support.
+// The Hermes service identifies what gRPCs Hermes will support.
 service Hermes {
 
   // Start monitoring a new storage system.
@@ -24,25 +24,23 @@ service Hermes {
   // Lists the storage systems being monitored at the moment.
   rpc ListMonitoredStorageSystems(ListMonitoredSystemsRequest) returns (ListMonitoredSystemsResponse) {}
 }
-// This is used for starting monitoring a new storage system using a Hermes probe.
+// HermesProbeRequest is used for starting monitoring a new storage system using a Hermes probe.
 message HermesProbeRequest {
   HermesProbeDef probe_config = 1;
 }
-// This is an empty response message for the probe response.
-// An exit status code may be added to probe responses, and included here.
+// TODO(#29) Add exit status to probe responses.
 message HermesProbeResponse {}
 
-// A target must be specified so that Hermes knows which target to stop monitoring.
+// StopMonitoringSystemRequest specifies a target so that Hermes knows which target to stop monitoring.
 message StopMonitoringSystemRequest {
   repeated TargetsDef target = 1;
 }
 // TODO(#29) Add exit status for probe responses.
 message StopMonitoringSystemResponse {}
 
-// Empty message for listing monitoring systems as no arguments needed.
 message ListMonitoredSystemsRequest {}
 
-// This message holds the list of targets returned by ListMonitoredStorageSystems()
+// ListMonitoredSystemsResponse message holds the list of targets returned by ListMonitoredStorageSystems()
 message ListMonitoredSystemsResponse {
   repeated TargetsDef target = 1;
 }
--- a/config/proto/service.proto
+++ b/config/proto/service.proto
@@ -1,6 +1,8 @@
 /*
   This proto defines the service-level config for Hermes.
   This is also the external service API for Hermes.
+  
+  Author: Evan Spendlove (@evanSpendlove)
 */
 syntax = "proto3";
 
@@ -12,7 +14,7 @@ import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
 option go_package = "github.com/googleinterns/step224-2020/config/proto";
 
 // This defines the (external) service-level config for Hermes.
-// The Hermes service identifies what gRPCs Hermes will support.
+// The Hermes service identifies the gRPCs Hermes supports.
 service Hermes {
 
   // Start monitoring a new storage system.
@@ -42,5 +44,6 @@ message ListMonitoredSystemsRequest {}
 
 // ListMonitoredSystemsResponse message holds the list of targets returned by ListMonitoredStorageSystems()
 message ListMonitoredSystemsResponse {
+  // TODO(#29) Add exit status to probe response.
   repeated TargetsDef target = 1;
 }
--- a/config/proto/service.proto
+++ b/config/proto/service.proto
@@ -1,3 +1,7 @@
+/*
+  This proto defines the service-level config for Hermes.
+  This is also the external service API for Hermes.
+*/
 syntax = "proto3";
 
 package hermes;
@@ -6,7 +10,8 @@ import "github.com/googleinterns/step224-2020/config/proto/probe.proto";
 import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
 
 option go_package = "github.com/googleinterns/step224-2020/config/proto";
-
+// This defines the (external) service-level config for Hermes.
+// It identifies what gRPCs Hermes will support.
 service Hermes {
 
   // Start monitoring a new storage system.
@@ -18,21 +23,26 @@ service Hermes {
   // Lists the storage systems being monitored at the moment.
   rpc ListMonitoredStorageSystems(ListMonitoredSystemsRequest) returns (ListMonitoredSystemsResponse) {}
 }
-
+// This is used for starting monitoring a new storage system using a Hermes probe.
 message HermesProbeRequest {
-  required HermesProbeDef probe_confg = 1;
+  required HermesProbeDef probe_config = 1;
 }
-
+// This is an empty response message for the probe response.
+// An exit status code may be added to probe responses, and included here.
 message HermesProbeResponse {}
 
+// A target must be specified so that Hermes knows which target to stop monitoring.
 message StopMonitoringSystemRequest {
   repeated targets.TargetsDef target = 1;
 }
-
+// This is an empty response message for the probe response.
+// An exit status code may be added to probe responses, and included here.
 message StopMonitoringSystemResponse {}
 
+// Empty message for listing monitoring systems as no arguments needed.
 message ListMonitoredSystemsRequest {}
 
+// This message holds the list of targets returned by ListMonitoredStorageSystems()
 message ListMonitoredSystemsResponse {
   repeated targets.TargetsDef target = 1;
 }
--- a/config/proto/service.proto
+++ b/config/proto/service.proto
@@ -25,7 +25,7 @@ service Hermes {
 }
 // This is used for starting monitoring a new storage system using a Hermes probe.
 message HermesProbeRequest {
-  required HermesProbeDef probe_config = 1;
+  HermesProbeDef probe_config = 1;
 }
 // This is an empty response message for the probe response.
 // An exit status code may be added to probe responses, and included here.
--- a/config/proto/service.proto
+++ b/config/proto/service.proto
@@ -25,7 +25,7 @@ message HermesProbeRequest {
 message HermesProbeResponse {}
 
 message StopMonitoringSystemRequest {
- required Target target = 1;
+  repeated targets.TargetsDef target = 1;
 }
 
 message StopMonitoringSystemResponse {}
@@ -33,5 +33,5 @@ message StopMonitoringSystemResponse {}
 message ListMonitoredSystemsRequest {}
 
 message ListMonitoredSystemsResponse {
-  repeated Target target = 1;
+  repeated targets.TargetsDef target = 1;
 }
--- a/config/proto/service.proto
+++ b/config/proto/service.proto
@@ -1,0 +1,37 @@
+syntax = "proto3";
+
+package hermes;
+
+import "github.com/googleinterns/step224-2020/config/proto/probe.proto";
+import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
+
+option go_package = "github.com/googleinterns/step224-2020/config/proto";
+
+service Hermes {
+
+  // Start monitoring a new storage system.
+  rpc StartMonitoringStorageSystem(HermesProbeRequest) returns (HermesProbeResponse) {}
+
+  rpc StopMonitoringStorageSystem(StopMonitoringSystemRequest) returns (StopMonitoringSystemResponse) {}
+
+  // Lists the storage systems being monitored at the moment.
+  rpc ListMonitoredStorageSystems(ListMonitoredSystemsRequest) returns (ListMonitoredSystemsResponse) {}
+}
+
+message HermesProbeRequest {
+  required HermesProbeDef probe_confg = 1;
+}
+
+message HermesProbeResponse {}
+
+message StopMonitoringSystemRequest {
+ required Target target = 1;
+}
+
+message StopMonitoringSystemResponse {}
+
+message ListMonitoredSystemsRequest {}
+
+message ListMonitoredSystemsResponse {
+  repeated Target target = 1;
+}
--- a/config/proto/targets.proto
+++ b/config/proto/targets.proto
@@ -4,6 +4,7 @@ package hermes;
 
 option go_package = "github.com/googleinterns/step224-2020/targets/proto";
 
+// Definition of a Target, a storage system to be probed by a specific probe.
 message TargetsDef {
   enum TargetSystem {
     GOOGLE_CLOUD_STORAGE = 0;
@@ -11,22 +12,29 @@ message TargetsDef {
   }
 
   enum ConnectionType {
-   HTTP = 0;
-   HTTPS = 1;
+     HTTP = 0;
+     HTTPS = 1;
   }
 
   required TargetSystem target_system = 1;
-  ConnectionType connection_type = 2;
+  required ConnectionType connection_type = 2;
   ConnectionDetails connection_details = 3;
-  required int64 total_space_allocated = 4;
+  required int64 total_space_allocated = 4; // in MBs.
 }
 
+// Necessary when connecting to an API (e.g. Ceph).
+// Unclear yet if we need this for GCS.
 message ConnectionDetails {
   string target_url = 1;
   int port = 2;
+  
+  // API key for the Ceph S3 API (if using Ceph) - irrelevant for GCS.
+  // API key can be gotten from your Ceph radosgw instance.
+  // Necessary for interacting with the Ceph S3 API.
   string api_key = 3;
 }
 
+// Allows the user to define multiple targets.
 message Targets {
   repeated TargetDef targets = 1;
 }
--- a/config/proto/targets.proto
+++ b/config/proto/targets.proto
@@ -1,6 +1,8 @@
 /*
   The Targets proto defines the config for a Target. 
   Targets are used by Hermes and Cloudprober to identify storage systems to monitor.
+  
+  Author: Evan Spendlove (@evanSpendlove)
 */
 syntax = "proto3";
 
@@ -17,7 +19,7 @@ message TargetsDef {
     GOOGLE_CLOUD_STORAGE = 0;
     CEPH = 1;
   }
-
+  // TODO(#30) Establish connection method for GCS and Ceph using Go libraries.
   enum ConnectionType {
      HTTP = 0;
      HTTPS = 1;
@@ -30,19 +32,17 @@ message TargetsDef {
   int64 total_space_allocated = 4;  // in MBs.
 }
 
-// Necessary when connecting to an API (e.g. Ceph).
-// Required for Ceph.
-// This functionality will be handled in the Go client library for GCS.
+// REQUIRED, when using Ceph as the target system.
+// Comparable functionality is handled in the Go client library for GCS.
 message CephConnectionDetails {
-  int32 port = 1;  // Port for connecting to Ceph API
+  int32 port = 1;  // Port for connecting to Ceph S3 API
   
-  // API key for the Ceph S3 API (if using Ceph) - irrelevant for GCS.
-  // API key can be gotten from your Ceph radosgw instance.
-  // Necessary for interacting with the Ceph S3 API.
+  // REQUIRED for Ceph S3 API
+  // GCS uses service account credentials instead.
   string api_key = 2;
 }
 
-// Allows the user to define multiple targets.
+// Targets allows the user to define multiple targets.
 message Targets {
   repeated TargetDef targets = 1;
 }
--- a/config/proto/targets.proto
+++ b/config/proto/targets.proto
@@ -8,7 +8,7 @@ package hermes;
 
 option go_package = "github.com/googleinterns/step224-2020/targets/proto";
 
-// Target contains all of the metadata necessary for Hermes to establish a connection to a storage system.
+// TargetsDef contains all of the metadata necessary for Hermes to establish a connection to a storage system.
 // Every probe request will require one or more targets.
 message TargetsDef {
   enum TargetSystem {
--- a/config/proto/targets.proto
+++ b/config/proto/targets.proto
@@ -25,8 +25,8 @@ message TargetsDef {
 // Necessary when connecting to an API (e.g. Ceph).
 // Unclear yet if we need this for GCS.
 message ConnectionDetails {
-  string target_url = 1;
-  int port = 2;
+  required string target_url = 1;
+  int port = 2; // Required for Ceph, but not for GCS since we will use the GCS go client library.
   
   // API key for the Ceph S3 API (if using Ceph) - irrelevant for GCS.
   // API key can be gotten from your Ceph radosgw instance.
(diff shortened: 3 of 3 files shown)


config/proto/probe.proto
```diff
@@ -0,0 +1,51 @@
+syntax = "proto3";
+
+package = hermes;
```
deniadelcid: Remove the =
```suggestion
package hermes;
```

config/proto/targets.proto
```diff
@@ -0,0 +1,32 @@
+syntax = "proto2";
+
+package hermes.targets
```
deniadelcid:
```suggestion
package hermes.targets;
```

config/proto/targets.proto
```diff
@@ -0,0 +1,32 @@
...
+
+option go_package = "github.com/googleinterns/hermes/targets/proto";
```
deniadelcid: Is it intentional to not include step224-2020 in the github path?

config/proto/targets.proto
```diff
@@ -0,0 +1,32 @@
...
+message TargetsDef {
+  enum TargetSystem {
+    GOOGLE_CLOUD_STORAGE = 0;
+    CEPH = 1;
+  }
+
+  enum ConnectionType {
+   HTTP = 0;
+   HTTPS = 1;
+  }
```
deniadelcid: Keep indentation consistent.

config/proto/service.proto
```diff
@@ -0,0 +1,37 @@
...
+message StopMonitoringSystemRequest {
+ required Target target = 1;
+}
+
+message StopMonitoringSystemResponse {}
+
+message ListMonitoredSystemsRequest {}
+
+message ListMonitoredSystemsResponse {
+  repeated Target target = 1;
```
evanSpendlove: Add targets.TargetDef as the type.
```suggestion
  repeated targets.TargetsDef target = 1;
```

config/proto/service.proto
```diff
@@ -0,0 +1,37 @@
...
+}
+
+message HermesProbeRequest {
+  required HermesProbeDef probe_confg = 1;
+}
+
+message HermesProbeResponse {}
+
+message StopMonitoringSystemRequest {
+ required Target target = 1;
```
evanSpendlove: Add targets.TargetDef as the type.
```suggestion
  repeated targets.TargetsDef target = 1;
```

config/proto/targets.proto
evanSpendlove: I updated this path in a later commit included in this pull request.

config/proto/targets.proto
```diff
@@ -0,0 +1,32 @@
...
+
+  required TargetSystem target_system = 1;
+  ConnectionType connection_type = 2;
+  ConnectionDetails connection_details = 3;
+  required int64 total_space_allocated = 4;
```
deniadelcid: Add comment. i.e. in which unit? bytes? MB? GiB?

config/proto/targets.proto
```diff
@@ -0,0 +1,32 @@
+syntax = "proto2";
+
+package hermes.targets
+
+option go_package = "github.com/googleinterns/hermes/targets/proto";
+
+message TargetsDef {
```
deniadelcid, line 15: Add comment at the top describing what the intent of the message

config/proto/targets.proto
```diff
@@ -0,0 +1,32 @@
...
+   HTTPS = 1;
+  }
+
+  required TargetSystem target_system = 1;
+  ConnectionType connection_type = 2;
+  ConnectionDetails connection_details = 3;
+  required int64 total_space_allocated = 4;
+}
+
+message ConnectionDetails {
```
deniadelcid: Add comment specifying when this is needed

config/proto/targets.proto
```diff
@@ -0,0 +1,32 @@
...
+  string target_url = 1;
+  int port = 2;
+  string api_key = 3;
+}
+
+message Targets {
```
deniadelcid, line 46: Add comment at the top.

config/proto/targets.proto
```diff
@@ -0,0 +1,32 @@
...
+    CEPH = 1;
+  }
+
+  enum ConnectionType {
+   HTTP = 0;
+   HTTPS = 1;
+  }
+
+  required TargetSystem target_system = 1;
+  ConnectionType connection_type = 2;
```
deniadelcid, line 29: If one isn't provided, what would the default be? Add comment in the line right above it.

config/proto/targets.proto
```diff
@@ -0,0 +1,32 @@
...
+  ConnectionDetails connection_details = 3;
+  required int64 total_space_allocated = 4;
+}
+
+message ConnectionDetails {
+  string target_url = 1;
+  int port = 2;
+  string api_key = 3;
```
deniadelcid: Which API? Where can I (the user) get this key from? When will it be needed? Add comments above.

config/proto/service.proto
deniadelcid: I gather you are importing this from the targets.proto config file, make sure to prepend the 'Target' message type with its package path.

Ditto for the other mentions below.

config/proto/targets.proto
evanSpendlove: Okay:
```suggestion
// Definition of a Target, a storage system to be probed by a specific probe.
message TargetsDef {
```

config/proto/targets.proto
evanSpendlove:
```suggestion
  enum TargetSystem {
    GOOGLE_CLOUD_STORAGE = 0;
    CEPH = 1;
  }

  enum ConnectionType {
     HTTP = 0;
     HTTPS = 1;
  }
```

config/proto/service.proto
```diff
@@ -0,0 +1,37 @@
...
+}
+
+message StopMonitoringSystemResponse {}
```
deniadelcid, line 41: You may want to add some sort of status code as a message field. But you can leave this for future work, up to you. (same for the rest of the response messages.

You could do something like what they exemplify here [1] or here [2].

[1] https://developers.google.com/protocol-buffers/docs/proto3#other
[2] https://developers.google.com/protocol-buffers/docs/proto3#any

config/proto/targets.proto
evanSpendlove: It would default to HTTP, I think, but it makes more sense to make it required.
```suggestion
  required ConnectionType connection_type = 2;
```

config/proto/targets.proto
evanSpendlove: Sorry about that, I forgot to copy the comment across.
```suggestion
  required int64 total_space_allocated = 4; // in MBs.
```

config/proto/targets.proto
evanSpendlove: Added suggestion:
```suggestion
  
  // API key for the Ceph S3 API (if using Ceph) - irrelevant for GCS.
  // API key can be gotten from your Ceph radosgw instance.
  // Necessary for interacting with the Ceph S3 API.
  string api_key = 3;
```

config/proto/targets.proto
evanSpendlove:
```suggestion
// Allows the user to define multiple targets.
message Targets {
```

config/proto/targets.proto
evanSpendlove:
```suggestion
// Necessary when connecting to an API (e.g. Ceph).
// Unclear yet if we need this for GCS.
message ConnectionDetails {
```

config/proto/service.proto
```diff
@@ -0,0 +1,37 @@
...
+import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
+
+option go_package = "github.com/googleinterns/step224-2020/config/proto";
+
+service Hermes {
+
+  // Start monitoring a new storage system.
+  rpc StartMonitoringStorageSystem(HermesProbeRequest) returns (HermesProbeResponse) {}
+
+  rpc StopMonitoringStorageSystem(StopMonitoringSystemRequest) returns (StopMonitoringSystemResponse) {}
```
deniadelcid, line 24: nit: Despite the name of the RPC being pretty self explanatory, please add a comment in the line right above.

config/proto/service.proto
evanSpendlove: I have created a new issue, #16, which describes this. I think it would be better to implement this in the future when I know more about what to expect from the return.

config/proto/service.proto
evanSpendlove:
```suggestion
  // Stop monitoring a storage system that is currently being monitored.
  rpc StopMonitoringStorageSystem(StopMonitoringSystemRequest) returns (StopMonitoringSystemResponse) {}
```

config/proto/service.proto
evanSpendlove: I have addressed this in a new commit.

config/proto/probe.proto
```diff
@@ -0,0 +1,51 @@
...
+message HermesProbeDef {
+  required string probe_name= 1;
+  required targets.TargetsDef targets = 2;
+
+  enum TargetSystem {
+    GCS = 1;
+    CEPH = 2;
+  }
+
+  // Type of Probe
```
deniadelcid: Can you make the comment a bit more thorough, i.e. Probe types currently supported. Then in the lines right next to to each enum field, you add a one liner explanation for what it does.

config/proto/probe.proto
```diff
@@ -0,0 +1,51 @@
...
+  // Add as key-value pairs
+  repeated AdditionalLabel probe_latency_additional_label = 10;
+  repeated AdditionalLabel api_call_latency_additional_label = 11;
+}
+
+// Simple Key-Value pair for additional metric labels
+message AdditionalLabel {
+  required string key = 1;
+  required string value = 2;
+}
```
deniadelcid: Since this is a key-value pair, you could consider using a map [1] instead of AdditionalField. Just keep in mind that maps cannot have repeated, but the map syntax should be equivalent to what your use of AdditionalLabel here.

[1] https://developers.google.com/protocol-buffers/docs/proto#maps

config/proto/targets.proto
```diff
@@ -0,0 +1,32 @@
...
+
+  required TargetSystem target_system = 1;
+  ConnectionType connection_type = 2;
+  ConnectionDetails connection_details = 3;
+  required int64 total_space_allocated = 4;
+}
+
+message ConnectionDetails {
+  string target_url = 1;
+  int port = 2;
```
deniadelcid: Since this isn't required, add a comment specifying what the default port is/will be if one isn't provided.

deniadelcid (changes requested):
Looks  great over all. Left a few comments and nits.

To dos:
- [x] Use protoc to create the pb.go files
- [x] Add more comments. Especially at the top of every file briefly (~1-2 lines) describing the purpose of the messages contained in that file.

-- Reply --
I have added more comments to all of the files.
I can't compile the pb.go files until the PR is accepted as I need to import some of these files from github.
I've added an issue #20 to create the compiled files and will do so when the PR is completed.

config/proto/targets.proto
```diff
@@ -4,29 +4,37 @@ package hermes;
...
   enum ConnectionType {
-   HTTP = 0;
-   HTTPS = 1;
+     HTTP = 0;
+     HTTPS = 1;
   }
 
   required TargetSystem target_system = 1;
-  ConnectionType connection_type = 2;
+  required ConnectionType connection_type = 2;
```
deniadelcid: Also, apparently, required fields are things to avoid when using proto2 [1]. ¯\_(ツ)_/¯

[1] https://developers.google.com/protocol-buffers/docs/style#things_to_avoid

config/proto/probe.proto
```diff
@@ -0,0 +1,51 @@
...
+  ProbeType probe_type = 4;
+
+  optional int32 interval_sec = 5 [default = 3600]; // Probing interval in seconds
+
+  // Probes will not retry, they will fail if they timeout.
+  optional int32 timeout_sec = 6 [default = 60]; // Timeout in seconds
+
+  // If specified, latency is stored as a distribution
+  optional metrics.Dist probe_latency_distribution = 7;
+  optional metrics.Dist api_call_latency_distribution = 8;
```
deniadelcid: For a user it might not be clear what the difference between these two, please add some comments.

config/proto/targets.proto
evanSpendlove: I altered this package in a newer commit to make it:
package hermes;

config/proto/probe.proto
evanSpendlove: We will probably have several additional labels for the distribution metrics, so we need to be able to use `repeated` here.

I might extract this into a new metric proto though and add the majority of the shared labels directly to the latency distributions.

config/proto/probe.proto
evanSpendlove:
```suggestion
  // Probe types currently supported
```

config/proto/probe.proto
```diff
@@ -0,0 +1,51 @@
...
+  enum TargetSystem {
+    GCS = 1;
+    CEPH = 2;
+  }
+
+  // Type of Probe
+  enum ProbeType {
+    WRITE = 1;
+    DELETE = 2;
+    CHECK = 3;
```
evanSpendlove:
```suggestion
    GET = 1; // GET probe is for getting an object (file) from a storage system.
    PUT = 2; // PUT probe is for putting an object (file) into a storage system.  
    DELETE = 3; // DELETE probe is for deleting an object (file) from a storage system.
```

config/proto/probe.proto
evanSpendlove: Also added suggestion below - added comments to all lines of ProbeType.

config/proto/probe.proto
evanSpendlove:
```suggestion
  optional metrics.Dist probe_latency_distribution = 7; // This measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
  optional metrics.Dist api_call_latency_distribution = 8; // This measures the latency of the API calls for the relevant storage system targeted.
```

config/proto/targets.proto
evanSpendlove:
```suggestion
  required string target_url = 1;
  int port = 2; // Required for Ceph, but not for GCS since we will use the GCS go client library.
  string api_key = 3; // Required for Ceph, not for GCS which uses service account credentials in an environment variable.
```

config/proto/targets.proto
evanSpendlove: Added in a comment below across multiple lines.

config/proto/targets.proto
```diff
@@ -0,0 +1,40 @@
...
+  ConnectionDetails connection_details = 3;
+  required int64 total_space_allocated = 4; // in MBs.
+}
+
+// Necessary when connecting to an API (e.g. Ceph).
+// Unclear yet if we need this for GCS.
+message ConnectionDetails {
+  string target_url = 1;
+  int port = 2;
```
evanSpendlove:
```suggestion
  required string target_url = 1;
  int port = 2; // Required for Ceph, but not for GCS since we will use the GCS go client library.
```

config/proto/targets.proto
evanSpendlove: I think we might switch to Proto3 (dependent on confirmation with Alicja), which does not support required, so I'll review these when confirmed.

config/proto/probe.proto
```diff
@@ -0,0 +1,51 @@
+syntax = "proto3";
```
evanSpendlove, line 7:
```suggestion
/*
  This proto defines the initial probe interface for Hermes probes.
  It is used in a variety of other files as it is the top-level probe interface.
*/
syntax = "proto3";
```

config/proto/probe.proto
```diff
@@ -0,0 +1,51 @@
...
+
+package hermes;
+
+import "github.com/google/cloudprober/metrics/proto/dist.proto";
+import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
+
+option go_package = "github.com/googleinterns/step224-2020/config/proto";
+
```
evanSpendlove, line 16:
```suggestion
  // This defines the top-level probe interface for Hermes probes.
  // All Hermes probes should extend this probe config.
```

config/proto/probe.proto
```diff
@@ -0,0 +1,51 @@
...
+  // Valid values: "ns", "us" (or "µs"), "ms", "s", "m", "h".
+  optional string latency_unit = 9 [default = "us"];
+
+  // Additional labels for the probe latency and API call latency distribution metrics.
+  // Add as key-value pairs
+  repeated AdditionalLabel probe_latency_additional_label = 10;
+  repeated AdditionalLabel api_call_latency_additional_label = 11;
+}
+
+// Simple Key-Value pair for additional metric labels
```
evanSpendlove:
```suggestion
// Simple Key-Value pair for additional metric labels
// Uses key, value instead of Map to allow for `repeated` keyword.
```

config/proto/service.proto
```diff
@@ -0,0 +1,38 @@
+syntax = "proto3";
```
evanSpendlove, line 7:
```suggestion
/*
  This proto defines the service-level config for Hermes.
  This is also the external service API for Hermes.
*/
syntax = "proto3";
```

config/proto/service.proto
```diff
@@ -0,0 +1,38 @@
...
+
+package hermes;
+
+import "github.com/googleinterns/step224-2020/config/proto/probe.proto";
+import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
+
+option go_package = "github.com/googleinterns/step224-2020/config/proto";
+
```
evanSpendlove, line 15:
```suggestion
// This defines the (external) service-level config for Hermes.
// It identifies what gRPCs Hermes will support.
```

config/proto/service.proto
```diff
@@ -0,0 +1,38 @@
...
+  // Start monitoring a new storage system.
+  rpc StartMonitoringStorageSystem(HermesProbeRequest) returns (HermesProbeResponse) {}
+
+  // Stop monitoring a storage system that is currently being monitored.
+  rpc StopMonitoringStorageSystem(StopMonitoringSystemRequest) returns (StopMonitoringSystemResponse) {}
+
+  // Lists the storage systems being monitored at the moment.
+  rpc ListMonitoredStorageSystems(ListMonitoredSystemsRequest) returns (ListMonitoredSystemsResponse) {}
+}
+
```
evanSpendlove:
```suggestion
// This is used for starting monitoring a new storage system using a Hermes probe.
```

config/proto/service.proto
```diff
@@ -0,0 +1,38 @@
...
+message HermesProbeRequest {
+  required HermesProbeDef probe_confg = 1;
```
evanSpendlove:
```suggestion
  required HermesProbeDef probe_config = 1;
```

config/proto/service.proto
```diff
@@ -0,0 +1,38 @@
...
+}
+
```
evanSpendlove:
```suggestion
// This is an empty response message for the probe response.
// An exit status code may be added to probe responses, and included here.
```

config/proto/service.proto
```diff
@@ -0,0 +1,38 @@
...
+message HermesProbeResponse {}
+
+message StopMonitoringSystemRequest {
```
evanSpendlove, line 37:
```suggestion
// A target must be specified so that Hermes knows which target to stop monitoring.
message StopMonitoringSystemRequest {
```

config/proto/service.proto
```diff
@@ -0,0 +1,38 @@
...
+  repeated targets.TargetsDef target = 1;
+}
+
```
evanSpendlove:
```suggestion
// This is an empty response message for the probe response.
// An exit status code may be added to probe responses, and included here.
```

config/proto/service.proto
```diff
@@ -0,0 +1,38 @@
...
+message StopMonitoringSystemResponse {}
+
```
evanSpendlove, line 42:
```suggestion

// Empty message for listing monitoring systems as no arguments needed.
```

config/proto/service.proto
```diff
@@ -0,0 +1,38 @@
...
+message ListMonitoredSystemsRequest {}
+
+message ListMonitoredSystemsResponse {
```
evanSpendlove, line 46:
```suggestion
// This message holds the list of targets returned by ListMonitoredStorageSystems()
message ListMonitoredSystemsResponse {
```

config/proto/targets.proto
```diff
@@ -0,0 +1,40 @@
+syntax = "proto2";
```
evanSpendlove:
```suggestion
/*
  This proto defines the config for a Target. 
  Targets are used by Hermes and Cloudprober to identify storage systems to monitor.
*/
syntax = "proto2";
```

config/proto/targets.proto
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  required TargetSystem target_system = 1;
+  required ConnectionType connection_type = 2;
+  ConnectionDetails connection_details = 3;
+  required int64 total_space_allocated = 4; // in MBs.
+}
+
+// Necessary when connecting to an API (e.g. Ceph).
+// Unclear yet if we need this for GCS.
```
evanSpendlove:
```suggestion
// Unclear yet if we need this for GCS, as we will use the Go client library and service account credentials.
```

alicjakwie (approved):
Everything already addressed by Denia

config/proto/probe.proto
```diff
@@ -0,0 +1,57 @@
...
+package hermes;
+
+import "github.com/google/cloudprober/metrics/proto/dist.proto";
+import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
+
+option go_package = "github.com/googleinterns/step224-2020/config/proto";
+  // This defines the top-level probe interface for Hermes probes.
+  // All Hermes probes should extend this probe config.
+message HermesProbeDef {
+  required string probe_name= 1;
```
evanSpendlove: Remove required for proto3.

```suggestion
  string probe_name= 1;
```

config/proto/probe.proto
```diff
@@ -0,0 +1,57 @@
...
+  required targets.TargetsDef targets = 2;
```
evanSpendlove:
```suggestion
  targets.TargetsDef targets = 2;
```

config/proto/probe.proto
```diff
@@ -0,0 +1,57 @@
...
+  enum ProbeType {
+    GET = 1; // GET probe is for getting an object (file) from a storage system.
+    PUT = 2; // PUT probe is for putting an object (file) into a storage system.  
+    DELETE = 3; // DELETE probe is for deleting an object (file) from a storage system.
+  }
+
+  TargetSystem target_system = 3;
+  ProbeType probe_type = 4;
+
+  optional int32 interval_sec = 5 [default = 3600]; // Probing interval in seconds
```
evanSpendlove:
```suggestion
  int32 interval_sec = 5 [default = 3600]; // Probing interval in seconds
```

config/proto/probe.proto
```diff
@@ -0,0 +1,57 @@
...
+
+  // Probes will not retry, they will fail if they timeout.
+  optional int32 timeout_sec = 6 [default = 60]; // Timeout in seconds
```
evanSpendlove:
```suggestion
  int32 timeout_sec = 6 [default = 60]; // Timeout in seconds
```

config/proto/probe.proto
```diff
@@ -0,0 +1,57 @@
...
+
+  // If specified, latency is stored as a distribution
+  optional metrics.Dist probe_latency_distribution = 7; // This measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
+  optional metrics.Dist api_call_latency_distribution = 8; // This measures the latency of the API calls for the relevant storage system targeted.
```
evanSpendlove:
```suggestion
  metrics.Dist probe_latency_distribution = 7; // This measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
  metrics.Dist api_call_latency_distribution = 8; // This measures the latency of the API calls for the relevant storage system targeted.
```

config/proto/probe.proto
```diff
@@ -0,0 +1,57 @@
...
+
+  // Valid values: "ns", "us" (or "µs"), "ms", "s", "m", "h".
+  optional string latency_unit = 9 [default = "us"];
```
evanSpendlove:
```suggestion
  string latency_unit = 9 [default = "us"];
```

config/proto/probe.proto
```diff
@@ -0,0 +1,57 @@
...
+  // Add as key-value pairs
+  repeated AdditionalLabel probe_latency_additional_label = 10;
+  repeated AdditionalLabel api_call_latency_additional_label = 11;
+}
+
+// Simple Key-Value pair for additional metric labels
+// Uses key, value instead of Map to allow for `repeated` keyword.
+message AdditionalLabel {
+  required string key = 1;
+  required string value = 2;
```
evanSpendlove:
```suggestion
  string key = 1;
  string value = 2;
```

config/proto/service.proto
```diff
@@ -0,0 +1,48 @@
...
+
+  // Stop monitoring a storage system that is currently being monitored.
+  rpc StopMonitoringStorageSystem(StopMonitoringSystemRequest) returns (StopMonitoringSystemResponse) {}
+
+  // Lists the storage systems being monitored at the moment.
+  rpc ListMonitoredStorageSystems(ListMonitoredSystemsRequest) returns (ListMonitoredSystemsResponse) {}
+}
+// This is used for starting monitoring a new storage system using a Hermes probe.
+message HermesProbeRequest {
+  required HermesProbeDef probe_config = 1;
```
evanSpendlove:
```suggestion
  HermesProbeDef probe_config = 1;
```

config/proto/targets.proto
```diff
@@ -0,0 +1,44 @@
+/*
+  This proto defines the config for a Target. 
+  Targets are used by Hermes and Cloudprober to identify storage systems to monitor.
+*/
+syntax = "proto2";
```
evanSpendlove: Update syntax to proto3.
```suggestion
syntax = "proto3";
```

config/proto/targets.proto
```diff
@@ -0,0 +1,44 @@
...
+    CEPH = 1;
+  }
+
+  enum ConnectionType {
+     HTTP = 0;
+     HTTPS = 1;
+  }
+
+  required TargetSystem target_system = 1;
+  required ConnectionType connection_type = 2;
```
evanSpendlove:
```suggestion
  TargetSystem target_system = 1;
  ConnectionType connection_type = 2;
```

config/proto/targets.proto
```diff
@@ -0,0 +1,44 @@
...
+  ConnectionDetails connection_details = 3;
+  required int64 total_space_allocated = 4; // in MBs.
```
evanSpendlove:
```suggestion
  int64 total_space_allocated = 4; // in MBs.
```

config/proto/targets.proto
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// Necessary when connecting to an API (e.g. Ceph).
+// Unclear yet if we need this for GCS, as we will use the Go client library and service account credentials.
+message ConnectionDetails {
+  required string target_url = 1;
```
evanSpendlove:
```suggestion
  string target_url = 1;
```

config/proto/targets.proto
```diff
@@ -0,0 +1,44 @@
...
+  int port = 2; // Required for Ceph, but not for GCS since we will use the GCS go client library.
```
evanSpendlove:
```suggestion
  int32 port = 2; // Required for Ceph, but not for GCS since we will use the GCS go client library.
```

config/proto/targets.proto
evanSpendlove: We are switching to Proto3!

evanSpendlove (commented):
Changes for making these files proto3 compliant.

config/proto/probe.proto
deniadelcid: A map by definition can hold many key-value pairs. As I said using a single map, you achieve the same result as having repeated key-value messages.

config/proto/probe.proto
```diff
@@ -0,0 +1,57 @@
...
+  // If specified, latency is stored as a distribution
+  metrics.Dist probe_latency_distribution = 7; // This measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
+  metrics.Dist api_call_latency_distribution = 8; // This measures the latency of the API calls for the relevant storage system targeted.
+
+  // Valid values: "ns", "us" (or "µs"), "ms", "s", "m", "h".
+  string latency_unit = 9 [default = "us"];
+
+  // Additional labels for the probe latency and API call latency distribution metrics.
+  // Add as key-value pairs
+  repeated AdditionalLabel probe_latency_additional_label = 10;
```
evanSpendlove, line 48:
```suggestion
  map<string, string> probe_latency_additional_label = 10;
```

config/proto/probe.proto
```diff
@@ -0,0 +1,57 @@
...
+  repeated AdditionalLabel api_call_latency_additional_label = 11;
```
evanSpendlove, line 49:
```suggestion
  map<string, string> api_call_latency_additional_label = 11;
```

config/proto/probe.proto
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// Simple Key-Value pair for additional metric labels
+// Uses key, value instead of Map to allow for `repeated` keyword.
+message AdditionalLabel {
+  string key = 1;
+  string value = 2;
+}
```
evanSpendlove, line 57: Replaced with Map.

```suggestion
```

config/proto/probe.proto
evanSpendlove: Yes, of course.. I think I forgot how maps work for a minute.. I think that a map would be a much cleaner solution, but using repeated AdditionalLabels should be more compatible with Cloudprober's configs, so when we implement a config that extends the Cloudprober probe config, we don't have duplicate fields (it uses AdditionalLabel too).

config/proto/probe.proto
```diff
@@ -13,8 +13,8 @@ option go_package = "github.com/googleinterns/step224-2020/config/proto";
   // This defines the top-level probe interface for Hermes probes.
   // All Hermes probes should extend this probe config.
 message HermesProbeDef {
-  required string probe_name= 1;
-  required targets.TargetsDef targets = 2;
+  string probe_name= 1;
+  targets.TargetsDef targets = 2;
```
deniadelcid: Remove. TargetsDef is now part of the "hermes" package, so it isn't necessary to include a path any more.

config/proto/probe.proto
```diff
@@ -13,8 +13,8 @@ option go_package = "github.com/googleinterns/step224-2020/config/proto";
...
 
   enum TargetSystem {
```
deniadelcid, line 25: Add comment above indicating that you expect this to be a storage system and that you're expecting it to stay constant until support for new storage systems is added.

config/proto/probe.proto
```diff
@@ -31,17 +31,17 @@ message HermesProbeDef {
   TargetSystem target_system = 3;
   ProbeType probe_type = 4;
 
-  optional int32 interval_sec = 5 [default = 3600]; // Probing interval in seconds
+  int32 interval_sec = 5 [default = 3600]; // Probing interval in seconds
```
deniadelcid: Leave two spaces after the ; for incline comments.
```suggestion
  int32 interval_sec = 5 [default = 3600];  // Probing interval in seconds
```

config/proto/probe.proto
```diff
@@ -52,6 +52,6 @@ message HermesProbeDef {
 // Simple Key-Value pair for additional metric labels
 // Uses key, value instead of Map to allow for `repeated` keyword.
 message AdditionalLabel {
```
deniadelcid: As mentioned before, change this to a map, the effect is the same. With a map, you could just append more entries if you wanted to add new key-value pairs, which is why they don't support repeated fields.

For instance, you could have a single AdditionalLabels map with two entries, one for probe  latency and the other for API call. Or if you believe these should be a different maps, you can then just append key-value pairs to each map instead.

config/proto/service.proto
```diff
@@ -0,0 +1,48 @@
...
+*/
+syntax = "proto3";
+
+package hermes;
+
+import "github.com/googleinterns/step224-2020/config/proto/probe.proto";
+import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
+
+option go_package = "github.com/googleinterns/step224-2020/config/proto";
+// This defines the (external) service-level config for Hermes.
```
deniadelcid, line 16: Add a blank line here.
```suggestion

// This defines the (external) service-level config for Hermes.
```

config/proto/service.proto
```diff
@@ -0,0 +1,48 @@
...
+// It identifies what gRPCs Hermes will support.
```
deniadelcid: Don't use pronouns at the beginning of a comment, reference the service/message/function/etc. name itself.
```suggestion
// Hermes identifies what gRPCs Hermes will support.
```

config/proto/service.proto
deniadelcid: You define target here as a repeated field, so make the comment reflect this by mentioning that at least one targets must be provided. Also what happens if no targets are provided as input? What about many targets? In case of several targets, will they be processed in order?

It's alright not to know now, but just keep in mind that these should probably be updated in the future when you implement these or make use of them in Go.

config/proto/service.proto
```diff
@@ -0,0 +1,48 @@
...
+message HermesProbeRequest {
+  HermesProbeDef probe_config = 1;
+}
+// This is an empty response message for the probe response.
+// An exit status code may be added to probe responses, and included here.
+message HermesProbeResponse {}
+
+// A target must be specified so that Hermes knows which target to stop monitoring.
+message StopMonitoringSystemRequest {
+  repeated targets.TargetsDef target = 1;
```
deniadelcid: Ditto here and in the rest all other TargetsDef mentions.
```suggestion
  repeated TargetsDef target = 1;
```

config/proto/service.proto
```diff
@@ -0,0 +1,48 @@
...
+}
+// This is an empty response message for the probe response.
+// An exit status code may be added to probe responses, and included here.
```
deniadelcid: It is unnecessary to mention that it's empty. Keep the comments consistent with what the system looks like right now. If you are committed to adding features/updates in the future, use TODO tags [1], to link to a corresponding/relevant Issue and brief instructions on what the todo here will be: 
e.g. // TODO(<link to issue regarding adding start/stop monitoring feature to hermes>): Add exit status for probe responses.

Another way of doing this is adding your github username instead of the bugname:
e.g. // TODO(username): Add exit status for probe responses.

config/proto/targets.proto
```diff
@@ -0,0 +1,44 @@
...
+  This proto defines the config for a Target. 
+  Targets are used by Hermes and Cloudprober to identify storage systems to monitor.
+*/
+syntax = "proto3";
+
+package hermes;
+
+option go_package = "github.com/googleinterns/step224-2020/targets/proto";
+
+// Definition of a Target, a storage system to be probed by a specific probe.
```
deniadelcid: This sounds a bit confusing. Maybe something along the lines of: Target contains all the metadata necessary for Hermes to establish a connection to a storage system. Every probe request will require a (at least one) target.

config/proto/targets.proto
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  TargetSystem target_system = 1;
+  ConnectionType connection_type = 2;
+  ConnectionDetails connection_details = 3;
+  int64 total_space_allocated = 4; // in MBs.
+}
+
+// Necessary when connecting to an API (e.g. Ceph).
+// Unclear yet if we need this for GCS, as we will use the Go client library and service account credentials.
```
deniadelcid: If it is unclear this will be needed for GCS then only make it required when the connection type is Ceph, so just make a comment on it. And just mention that this functionality will be handled in the Go client library/binary for GCS. If that changes, then just come back here when it does and update the comment.

config/proto/targets.proto
```diff
@@ -0,0 +1,44 @@
...
+message ConnectionDetails {
```
deniadelcid: If most of these configs will only be relevant for Ceph, then take the target URL out and make it part of the TargetsDef and then rename this message to something Ceph specific (e.g. CephConnectionDetails), then make the CephConnectionDetails an optional (in proto3 syntax/style friendly) field.

deniadelcid (changes requested):

config/proto/probe.proto
evanSpendlove:
```suggestion
  TargetsDef targets = 2;
```

config/proto/probe.proto
evanSpendlove:
```suggestion
  // TargetSystem is expected to be a storage system.
  // It will stay constant until support for new storage systems is added.
  enum TargetSystem {
```

config/proto/probe.proto
```diff
@@ -0,0 +1,57 @@
+/*
+  This proto defines the initial probe interface for Hermes probes.
+  It is used in a variety of other files as it is the top-level probe interface.
+*/
+syntax = "proto3";
+
+package hermes;
+
```
evanSpendlove, line 10:
```suggestion

import "github.com/google/cloudprober/probes/proto/config.proto";
```

config/proto/probe.proto
evanSpendlove:
```suggestion
  repeated cloudprober.probes.AdditionalLabel probe_latency_additional_label = 10;
  repeated cloudprober.probes.AdditionalLabel api_call_latency_additional_label = 11;
```

config/proto/probe.proto
evanSpendlove: I have imported the Cloudprober proto and used `cloudprober.probes.AdditionalLabel` as we discussed.

config/proto/probe.proto
evanSpendlove:
```suggestion
```

config/proto/service.proto
```diff
@@ -0,0 +1,48 @@
...
+message StopMonitoringSystemResponse {}
+
+// Empty message for listing monitoring systems as no arguments needed.
+message ListMonitoredSystemsRequest {}
+
+// This message holds the list of targets returned by ListMonitoredStorageSystems()
+message ListMonitoredSystemsResponse {
+  repeated targets.TargetsDef target = 1;
```
evanSpendlove:
```suggestion
  repeated TargetsDef target = 1;
```

config/proto/targets.proto
evanSpendlove:
```suggestion
// Required for Ceph.
// This functionality will be handled in the Go client library for GCS.
```

config/proto/probe.proto
```diff
@@ -0,0 +1,57 @@
...
+import "github.com/google/cloudprober/metrics/proto/dist.proto";
+import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
+
+option go_package = "github.com/googleinterns/step224-2020/config/proto";
+  // This defines the top-level probe interface for Hermes probes.
+  // All Hermes probes should extend this probe config.
```
evanSpendlove:
```suggestion

// This defines the top-level probe interface for Hermes probes.
// All Hermes probes should extend this probe config.
```

config/proto/probe.proto
```diff
@@ -0,0 +1,57 @@
...
+  enum TargetSystem {
+    GCS = 1;
+    CEPH = 2;
+  }
+
+  // Probe types currently supported
+  enum ProbeType {
+    GET = 1; // GET probe is for getting an object (file) from a storage system.
+    PUT = 2; // PUT probe is for putting an object (file) into a storage system.  
+    DELETE = 3; // DELETE probe is for deleting an object (file) from a storage system.
```
evanSpendlove:
```suggestion
    GET = 1;  // GET probe is for getting an object (file) from a storage system.
    PUT = 2;  // PUT probe is for putting an object (file) into a storage system.  
    DELETE = 3;  // DELETE probe is for deleting an object (file) from a storage system.
```

config/proto/probe.proto
```diff
@@ -0,0 +1,57 @@
...
+  ProbeType probe_type = 4;
+
+  int32 interval_sec = 5 [default = 3600]; // Probing interval in seconds
+
+  // Probes will not retry, they will fail if they timeout.
+  int32 timeout_sec = 6 [default = 60]; // Timeout in seconds
+
+  // If specified, latency is stored as a distribution
+  metrics.Dist probe_latency_distribution = 7; // This measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
+  metrics.Dist api_call_latency_distribution = 8; // This measures the latency of the API calls for the relevant storage system targeted.
```
evanSpendlove:
```suggestion
  int32 timeout_sec = 6 [default = 60];  // Timeout in seconds

  // If specified, latency is stored as a distribution
  metrics.Dist probe_latency_distribution = 7;  // This measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
  metrics.Dist api_call_latency_distribution = 8;  // This measures the latency of the API calls for the relevant storage system targeted.
```

config/proto/targets.proto
```diff
@@ -0,0 +1,44 @@
...
+
+  enum ConnectionType {
+     HTTP = 0;
+     HTTPS = 1;
+  }
+
+  TargetSystem target_system = 1;
+  ConnectionType connection_type = 2;
+  ConnectionDetails connection_details = 3;
+  int64 total_space_allocated = 4; // in MBs.
```
evanSpendlove, line 26:
```suggestion
  int64 total_space_allocated = 4;  // in MBs.
```

evanSpendlove:
```suggestion
  string target_url = 5;  // URL for connecting to the the API of the target storage system.
  int64 total_space_allocated = 4;  // in MBs.
```

config/proto/targets.proto
evanSpendlove: I have removed target_url and placed it in the TargetsDef message.

In proto3, all fields are optional (no enforcement), so there isn't any special syntax required.
```suggestion
message CephConnectionDetails {
```

config/proto/targets.proto
```diff
@@ -0,0 +1,44 @@
...
+// Necessary when connecting to an API (e.g. Ceph).
+// Unclear yet if we need this for GCS, as we will use the Go client library and service account credentials.
+message ConnectionDetails {
+  string target_url = 1;
+  int32 port = 2; // Required for Ceph, but not for GCS since we will use the GCS go client library.
+  
+  // API key for the Ceph S3 API (if using Ceph) - irrelevant for GCS.
+  // API key can be gotten from your Ceph radosgw instance.
+  // Necessary for interacting with the Ceph S3 API.
+  string api_key = 3;
```
evanSpendlove:
```suggestion
  int32 port = 1;  // Port for connecting to Ceph API
  
  // API key for the Ceph S3 API (if using Ceph) - irrelevant for GCS.
  // API key can be gotten from your Ceph radosgw instance.
  // Necessary for interacting with the Ceph S3 API.
  string api_key = 2;
```

config/proto/targets.proto
evanSpendlove:
```suggestion
// Target contains all of the metadata necessary for Hermes to establish a connection to a storage system.
// Every probe request will require one or more targets.
```

config/proto/service.proto
evanSpendlove: Amended.

config/proto/service.proto
evanSpendlove: Okay, thank you for pointing this out, I'll use TODOs in future.

```suggestion
// TODO(#29) Add exit status for probe responses.
```

config/proto/probe.proto
```diff
@@ -0,0 +1,54 @@
+/*
+  This proto defines the initial probe interface for Hermes probes.
+  It is used in a variety of other files as it is the top-level probe interface.
```
evanSpendlove, line 3:
```suggestion
  Probe defines the initial probe interface for Hermes probes.
  Probe is used in a variety of other files as it is the top-level probe interface.
```

config/proto/probe.proto
```diff
@@ -0,0 +1,54 @@
...
+
+package hermes;
+
+import "github.com/google/cloudprober/probes/proto/config.proto";
+import "github.com/google/cloudprober/metrics/proto/dist.proto";
+import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
+
+option go_package = "github.com/googleinterns/step224-2020/config/proto";
+
+// This defines the top-level probe interface for Hermes probes.
```
evanSpendlove, line 15: HermesProbeDef

config/proto/service.proto
```diff
@@ -0,0 +1,48 @@
+/*
+  This proto defines the service-level config for Hermes.
+  This is also the external service API for Hermes.
```
evanSpendlove, line 3:
```suggestion
  Service defines the service-level config for Hermes.
  Service is also the external service API for Hermes.
```

config/proto/probe.proto
evanSpendlove:
```suggestion
// HermesProbeDef defines the top-level probe interface for Hermes probes.
```

config/proto/service.proto
```diff
@@ -0,0 +1,48 @@
...
+
+package hermes;
+
+import "github.com/googleinterns/step224-2020/config/proto/probe.proto";
+import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
+
+option go_package = "github.com/googleinterns/step224-2020/config/proto";
+
+// This defines the (external) service-level config for Hermes.
+// Hermes identifies what gRPCs Hermes will support.
```
evanSpendlove:
```suggestion
// The Hermes service identifies what gRPCs Hermes will support.
```

config/proto/service.proto
```diff
@@ -0,0 +1,48 @@
...
+  // Start monitoring a new storage system.
+  rpc StartMonitoringStorageSystem(HermesProbeRequest) returns (HermesProbeResponse) {}
+
+  // Stop monitoring a storage system that is currently being monitored.
+  rpc StopMonitoringStorageSystem(StopMonitoringSystemRequest) returns (StopMonitoringSystemResponse) {}
+
+  // Lists the storage systems being monitored at the moment.
+  rpc ListMonitoredStorageSystems(ListMonitoredSystemsRequest) returns (ListMonitoredSystemsResponse) {}
+}
+// This is used for starting monitoring a new storage system using a Hermes probe.
```
evanSpendlove:
```suggestion
// HermesProbeRequest is used for starting monitoring a new storage system using a Hermes probe.
```

config/proto/service.proto
```diff
@@ -0,0 +1,48 @@
...
+message HermesProbeRequest {
+  HermesProbeDef probe_config = 1;
+}
+// This is an empty response message for the probe response.
+// An exit status code may be added to probe responses, and included here.
```
evanSpendlove:
```suggestion
// TODO(#29) Add exit status to probe responses.
```

config/proto/service.proto
```diff
@@ -0,0 +1,48 @@
...
+message StopMonitoringSystemRequest {
+  repeated TargetsDef target = 1;
+}
+// TODO(#29) Add exit status for probe responses.
+message StopMonitoringSystemResponse {}
+
+// Empty message for listing monitoring systems as no arguments needed.
+message ListMonitoredSystemsRequest {}
+
+// This message holds the list of targets returned by ListMonitoredStorageSystems()
```
evanSpendlove:
```suggestion
// ListMonitoredSystemsResponse message holds the list of targets returned by ListMonitoredStorageSystems()
```

config/proto/service.proto
```diff
@@ -0,0 +1,48 @@
...
+}
+// This is used for starting monitoring a new storage system using a Hermes probe.
+message HermesProbeRequest {
+  HermesProbeDef probe_config = 1;
+}
+// This is an empty response message for the probe response.
+// An exit status code may be added to probe responses, and included here.
+message HermesProbeResponse {}
+
+// A target must be specified so that Hermes knows which target to stop monitoring.
```
evanSpendlove:
```suggestion
// StopMonitoringSystemRequest specifies a target so that Hermes knows which target to stop monitoring.
```

config/proto/service.proto
```diff
@@ -0,0 +1,48 @@
...
+message StopMonitoringSystemRequest {
+  repeated TargetsDef target = 1;
+}
+// TODO(#29) Add exit status for probe responses.
+message StopMonitoringSystemResponse {}
+
+// Empty message for listing monitoring systems as no arguments needed.
```
evanSpendlove:
```suggestion
```

config/proto/targets.proto
```diff
@@ -0,0 +1,46 @@
...
+  This proto defines the config for a Target. 
+  Targets are used by Hermes and Cloudprober to identify storage systems to monitor.
+*/
+syntax = "proto3";
+
+package hermes;
+
+option go_package = "github.com/googleinterns/step224-2020/targets/proto";
+
+// Target contains all of the metadata necessary for Hermes to establish a connection to a storage system.
```
evanSpendlove:
```suggestion
// TargetsDef contains all of the metadata necessary for Hermes to establish a connection to a storage system.
```

config/proto/targets.proto
```diff
@@ -0,0 +1,46 @@
+/*
+  This proto defines the config for a Target.
```
evanSpendlove:
```suggestion
  The Targets proto defines the config for a Target. 
```

config/proto/targets.proto
```diff
@@ -0,0 +1,46 @@
...
+syntax = "proto3";
+
+package hermes;
+
+option go_package = "github.com/googleinterns/step224-2020/targets/proto";
+
+// TargetsDef contains all of the metadata necessary for Hermes to establish a connection to a storage system.
+// Every probe request will require one or more targets.
+message TargetsDef {
+  enum TargetSystem {
```
evanSpendlove, line 18:
```suggestion
  // TargetSystem is expected to be a storage system.
  // It will stay constant until support for new storage systems is added.
  enum TargetSystem {
```

config/proto/probe.proto
```diff
@@ -6,52 +6,49 @@ syntax = "proto3";
...
+  int32 interval_sec = 5 [default = 3600];  // Probing interval in seconds
 
   // Probes will not retry, they will fail if they timeout.
-  int32 timeout_sec = 6 [default = 60]; // Timeout in seconds
+  int32 timeout_sec = 6 [default = 60];  // Timeout in seconds
 
   // If specified, latency is stored as a distribution
-  metrics.Dist probe_latency_distribution = 7; // This measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
-  metrics.Dist api_call_latency_distribution = 8; // This measures the latency of the API calls for the relevant storage system targeted.
+  metrics.Dist probe_latency_distribution = 7;  // This measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
```
deniadelcid:
```suggestion
  metrics.Dist probe_latency_distribution = 7;  // Measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
```

config/proto/probe.proto
```diff
@@ -6,52 +6,49 @@ syntax = "proto3";
...
+  metrics.Dist api_call_latency_distribution = 8;  // This measures the latency of the API calls for the relevant storage system targeted.
```
deniadelcid:
```suggestion
  metrics.Dist api_call_latency_distribution = 8;  // Measures the latency of the API calls for the relevant storage system targeted.
```

config/proto/targets.proto
```diff
@@ -0,0 +1,48 @@
...
+  TargetSystem target_system = 1;
+  ConnectionType connection_type = 2;
+  ConnectionDetails connection_details = 3;
+  string target_url = 5;  // URL for connecting to the the API of the target storage system.
+  int64 total_space_allocated = 4;  // in MBs.
+}
+
+// Necessary when connecting to an API (e.g. Ceph).
+// Required for Ceph.
+// This functionality will be handled in the Go client library for GCS.
```
deniadelcid:
```suggestion
// REQUIRED, when using Ceph as the target system.
// Comparable functionality is handled in the Go client library for GCS.
```

config/proto/service.proto
```diff
@@ -0,0 +1,46 @@
...
+
+package hermes;
+
+import "github.com/googleinterns/step224-2020/config/proto/probe.proto";
+import "github.com/googleinterns/step224-2020/config/proto/targets.proto";
+
+option go_package = "github.com/googleinterns/step224-2020/config/proto";
+
+// This defines the (external) service-level config for Hermes.
+// The Hermes service identifies what gRPCs Hermes will support.
```
deniadelcid:
```suggestion
// The Hermes service identifies the gRPCs Hermes supports.
```

deniadelcid (commented):
Looking great!

Just a quick reminder to keep your comment in present tense rather than future, hypothetical tense. i.e. say things such as "Hermes supports ...", rather than "Hermes will support..."

-- Reply --
I've accepted your suggestions and made several minor grammatical changes in the comments across these files.
I've also added the 'author' section to the header comment in each file.

config/proto/probe.proto
evanSpendlove, line 3:
```suggestion
  It is used in a variety of other files as it is the top-level probe interface.
  
  Author: Evan Spendlove (@evanSpendlove)
```

config/proto/service.proto
```diff
@@ -0,0 +1,46 @@
+/*
+  This proto defines the service-level config for Hermes.
+  This is also the external service API for Hermes.
```
evanSpendlove, line 3:
```suggestion
  This is also the external service API for Hermes.
  
  Author: Evan Spendlove (@evanSpendlove)
```

config/proto/service.proto
```diff
@@ -0,0 +1,46 @@
...
+  repeated TargetsDef target = 1;
+}
+// TODO(#29) Add exit status for probe responses.
+message StopMonitoringSystemResponse {}
+
+message ListMonitoredSystemsRequest {}
+
+// ListMonitoredSystemsResponse message holds the list of targets returned by ListMonitoredStorageSystems()
+message ListMonitoredSystemsResponse {
+  repeated TargetsDef target = 1;
```
evanSpendlove, line 48:
```suggestion
  // TODO(#29) Add exit status to probe response.
  repeated TargetsDef target = 1;
```

config/proto/targets.proto
```diff
@@ -0,0 +1,48 @@
+/*
+  The Targets proto defines the config for a Target. 
+  Targets are used by Hermes and Cloudprober to identify storage systems to monitor.
```
evanSpendlove, line 3:
```suggestion
  Targets are used by Hermes and Cloudprober to identify storage systems to monitor.
  
  Author: Evan Spendlove (@evanSpendlove)
```

config/proto/targets.proto
```diff
@@ -0,0 +1,48 @@
...
+// TargetsDef contains all of the metadata necessary for Hermes to establish a connection to a storage system.
+// Every probe request will require one or more targets.
+message TargetsDef {
+  // TargetSystem is expected to be a storage system.
+  // It will stay constant until support for new storage systems is added.
+  enum TargetSystem {
+    GOOGLE_CLOUD_STORAGE = 0;
+    CEPH = 1;
+  }
+
```
evanSpendlove:
```suggestion
  // TODO(#30) Establish connection method for GCS and Ceph using Go libraries.
```

config/proto/targets.proto
```diff
@@ -0,0 +1,48 @@
...
+// Necessary when connecting to an API (e.g. Ceph).
+// Required for Ceph.
+// This functionality will be handled in the Go client library for GCS.
+message CephConnectionDetails {
+  int32 port = 1;  // Port for connecting to Ceph API
+  
+  // API key for the Ceph S3 API (if using Ceph) - irrelevant for GCS.
+  // API key can be gotten from your Ceph radosgw instance.
+  // Necessary for interacting with the Ceph S3 API.
```
evanSpendlove:
```suggestion
  // REQUIRED for Ceph S3 API
  // GCS uses service account credentials instead.
```

config/proto/targets.proto
```diff
@@ -0,0 +1,48 @@
...
+  string api_key = 2;
+}
+
+// Allows the user to define multiple targets.
```
evanSpendlove:
```suggestion
// Targets allows the user to define multiple targets.
```

config/proto/targets.proto
```diff
@@ -0,0 +1,48 @@
...
+  ConnectionDetails connection_details = 3;
+  string target_url = 5;  // URL for connecting to the the API of the target storage system.
+  int64 total_space_allocated = 4;  // in MBs.
+}
+
+// Necessary when connecting to an API (e.g. Ceph).
+// Required for Ceph.
+// This functionality will be handled in the Go client library for GCS.
+message CephConnectionDetails {
+  int32 port = 1;  // Port for connecting to Ceph API
```
evanSpendlove:
```suggestion
  int32 port = 1;  // Port for connecting to Ceph S3 API
```

deniadelcid (approved):

PR #40, opened 2020-08-25
Updated protos so they compile and added compiled protos.

evanSpendlove:


config/proto/probe.proto
```diff
@@ -1,9 +1,22 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Author: Evan Spendlove (@evanSpendlove)
+//
+// This proto defines the initial probe interface for Hermes probes.
```
deniadelcid:
```suggestion
// This proto defines the probe interface for Hermes probes.
```

config/proto/probe.proto
```diff
@@ -23,31 +36,33 @@ message HermesProbeDef {
...
   // It will stay constant until support for new storage systems is added.
   enum TargetSystem {
+    TARGET_SYSTEM_UNSPECIFIED = 0;
     GCS = 1;
     CEPH = 2;
   }
 
   // Probe types currently supported
   enum ProbeType {
+    PROBE_TYPE_UNSPECIFIED = 0; // For unspecified probe types
```
deniadelcid:
```suggestion
    PROBE_TYPE_UNSPECIFIED = 0;  // For unspecified probe types
```

deniadelcid: Which probes would fall under this category or under which circumstance(s) is it expected to send a probe of this type?

config/proto/probe.proto
```diff
@@ -23,31 +36,33 @@ message HermesProbeDef {
...
-    PUT = 2;  // PUT probe is for putting an object (file) into a storage system.  
+    PUT = 2;  // PUT probe is for putting an object (file) into a storage system.
     DELETE = 3;  // DELETE probe is for deleting an object (file) from a storage system.
   }
 
   TargetSystem target_system = 3;
   ProbeType probe_type = 4;
 
-  int32 interval_sec = 5 [default = 3600];  // Probing interval in seconds
+  int32 interval_sec = 5; // Probing interval in seconds
```
deniadelcid: If you plan on enforcing a default in the go implementation, mention this on the comment. Ditto for the ones below.

config/proto/targets.proto
```diff
@@ -27,22 +40,24 @@ message TargetsDef {
 
   TargetSystem target_system = 1;
   ConnectionType connection_type = 2;
-  ConnectionDetails connection_details = 3;
+  CephConnectionDetails ceph_connection_details = 3;
```
cianbgoog: Could you just make this a connection details? All the contents look like things that should be required for most probes, and. then you could reuse this for other storage systems in future?

config/proto/targets.proto
```diff
@@ -27,22 +40,24 @@ message TargetsDef {
...
+
   // REQUIRED for Ceph S3 API
   // GCS uses service account credentials instead.
   string api_key = 2;
 }
 
 // Targets allows the user to define multiple targets.
 message Targets {
-  repeated TargetDef targets = 1;
+  repeated TargetsDef targets = 1;
```
cianbgoog: Could you call this TargetDefinition? it's repeated, but it only includes one target, doesn't it?

cianbgoog:
It would also be really nice if this included a test, even if the test just checked that everything in it still compiles :-)

-- REPLY --
I've added a neat little bash script that does this 😄

config/proto/probe.proto
```diff
@@ -23,31 +36,33 @@ message HermesProbeDef {
...
 
   // If specified, latency is stored as a distribution
-  metrics.Dist probe_latency_distribution = 7;  // Measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
-  metrics.Dist api_call_latency_distribution = 8;  // Measures the latency of the API calls for the relevant storage system targeted.
+  cloudprober.metrics.Dist probe_latency_distribution = 7;  // Measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
+  cloudprober.metrics.Dist api_call_latency_distribution = 8;  // Measures the latency of the API calls for the relevant storage system targeted.
 
   // Valid values: "ns", "us" (or "µs"), "ms", "s", "m", "h".
-  string latency_unit = 9 [default = "us"];
+  string latency_unit = 9;
```
evanSpendlove:
```suggestion
  // Default: "us"
  string latency_unit = 9;
```

config/proto/probe.proto
evanSpendlove: I've added comments detailing the defaults for each field with a default value.

evanSpendlove:
```suggestion
  // Default = 3600
  int32 interval_sec = 5; // Probing interval in seconds
```

config/proto/probe.proto
```diff
@@ -23,31 +36,33 @@ message HermesProbeDef {
...
 
   TargetSystem target_system = 3;
   ProbeType probe_type = 4;
 
-  int32 interval_sec = 5 [default = 3600];  // Probing interval in seconds
+  int32 interval_sec = 5; // Probing interval in seconds
 
   // Probes will not retry, they will fail if they timeout.
-  int32 timeout_sec = 6 [default = 60];  // Timeout in seconds
+  int32 timeout_sec = 6;  // Timeout in seconds
```
evanSpendlove:
```suggestion
  // Default = 60
  int32 timeout_sec = 6;  // Timeout in seconds
```

config/proto/targets.proto
evanSpendlove:
```suggestion
  ConnectionDetails connection_details = 3;
```

config/proto/targets.proto
```diff
@@ -27,22 +40,24 @@ message TargetsDef {
...
   ConnectionType connection_type = 2;
-  ConnectionDetails connection_details = 3;
+  CephConnectionDetails ceph_connection_details = 3;
   string target_url = 5;  // URL for connecting to the the API of the target storage system.
   int64 total_space_allocated = 4;  // in MBs.
 }
 
 // REQUIRED, when using Ceph as the target system.
 // Comparable functionality is handled in the Go client library for GCS.
 message CephConnectionDetails {
```
evanSpendlove:
```suggestion
message ConnectionDetails {
```

config/proto/targets.proto
evanSpendlove: Yes, it only includes one target. I will change this.

evanSpendlove:
```suggestion
  repeated TargetDefinition targets = 1;
```

config/proto/probe.proto
cianbgoog: Could we make the default the default in the proto, rather than just as a comment?

(or is there a reason we've not done that?)

config/proto/targets.proto
evanSpendlove: Changed.

config/proto/probe.proto
evanSpendlove: None, ideally, however it is required that enums starts from 0 and it is recommended to use the 0 val for 'unspecified' values according to the Proto style guide.

Within the implementation, if the probe type is unspecified, an error will be thrown as this should not be the case.

Source: https://developers.google.com/protocol-buffers/docs/style#enums

config/proto/targets.proto
evanSpendlove: I've amended this to be ConnectionDetails. We had changed it to CephConnectionDetails as GCS won't need this, but it does make sense that other storage systems will require this in the future.

config/proto/targets.proto
```diff
@@ -34,15 +47,17 @@ message TargetsDef {
...
+
   // REQUIRED for Ceph S3 API
   // GCS uses service account credentials instead.
   string api_key = 2;
 }
 
 // Targets allows the user to define multiple targets.
 message Targets {
-  repeated TargetDef targets = 1;
+  repeated TargetDefinition targets = 1;
```
cianbgoog, line 62: You also need to change the definition of this message on line 28 (if I'm reading the github review properly)

config/proto/targets.proto
```diff
@@ -34,15 +47,17 @@ message TargetsDef {
 
 // REQUIRED, when using Ceph as the target system.
 // Comparable functionality is handled in the Go client library for GCS.
-message CephConnectionDetails {
-  int32 port = 1;  // Port for connecting to Ceph S3 API
-  
+message ConnectionDetails {
+  // REQUIRED for Ceph S3 API
```
cianbgoog, line 51: Stupid question maybe, but do you not need to give a hostname for ceph?

cianbgoog (changes requested):

config/proto/probe.proto
deniadelcid:
```suggestion
  string latency_unit = 9;  // default is "us"
```

config/proto/probe.proto
deniadelcid:
```suggestion
  int32 interval_sec = 5;  // Probing interval in seconds, default = 3600
```

deniadelcid: Proto3 doesn't support custom default values [1]. All the default values it has are null/zero equivalents depending on the field's data type [2], which can't be modified.

[1] https://cloud.google.com/apis/design/proto3
[2] https://developers.google.com/protocol-buffers/docs/proto3?hl=en#default

config/proto/probe.proto
```diff
@@ -52,16 +52,19 @@ message HermesProbeDef {
   TargetSystem target_system = 3;
   ProbeType probe_type = 4;
 
+  // Default = 3600
```
deniadelcid: Place these default comments as inline comments. They are harder to miss this way.

deniadelcid (approved):
Left few more minor comments, please fix before merging. But overall LGTM.

config/proto/probe.proto
```diff
@@ -18,36 +31,41 @@ option go_package = "github.com/googleinterns/step224-2020/config/proto";
...
   ProbeType probe_type = 4;
 
-  int32 interval_sec = 5 [default = 3600];  // Probing interval in seconds
+  // Default = 3600
+  int32 interval_sec = 5; // Probing interval in seconds
 
   // Probes will not retry, they will fail if they timeout.
-  int32 timeout_sec = 6 [default = 60];  // Timeout in seconds
+  // Default = 60
+  int32 timeout_sec = 6;  // Timeout in seconds
```
evanSpendlove:
```suggestion
  int32 timeout_sec = 6;  // Timeout in seconds, default = 60
```

config/proto/probe.proto
```diff
@@ -18,36 +31,41 @@ option go_package = "github.com/googleinterns/step224-2020/config/proto";
...
-    PUT = 2;  // PUT probe is for putting an object (file) into a storage system.  
+    PUT = 2;  // PUT probe is for putting an object (file) into a storage system.
     DELETE = 3;  // DELETE probe is for deleting an object (file) from a storage system.
   }
 
   TargetSystem target_system = 3;
   ProbeType probe_type = 4;
 
-  int32 interval_sec = 5 [default = 3600];  // Probing interval in seconds
+  // Default = 3600
```
evanSpendlove:
```suggestion
```

config/proto/targets.proto
evanSpendlove: I think this should be contained in the target_url for connecting to the API?

config/proto/targets.proto
evanSpendlove: Yup, I did that in a more recent commit as I needed to change it across multiple files.

evanSpendlove (commented):
Resolved some comments.

config/proto/probe.proto
```diff
@@ -18,36 +31,41 @@ option go_package = "github.com/googleinterns/step224-2020/config/proto";
...
 
   // If specified, latency is stored as a distribution
-  metrics.Dist probe_latency_distribution = 7;  // Measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
-  metrics.Dist api_call_latency_distribution = 8;  // Measures the latency of the API calls for the relevant storage system targeted.
+  cloudprober.metrics.Dist probe_latency_distribution = 7;  // Measures the latency of Hermes' probes, allowing you to diagnose a network issue or issue with Hermes.
+  cloudprober.metrics.Dist api_call_latency_distribution = 8;  // Measures the latency of the API calls for the relevant storage system targeted.
 
   // Valid values: "ns", "us" (or "µs"), "ms", "s", "m", "h".
-  string latency_unit = 9 [default = "us"];
+  // Default: "us"
```
evanSpendlove:
```suggestion
```

config/proto/targets.proto
cianbgoog: cual beans

cianbgoog (approved):

config/proto/probe.proto
evanSpendlove: I've done so across all files.

PR #42, opened 2020-08-27
Added go fmt to test.sh

evanSpendlove:
One change: added `go fmt` to the test.sh script, so that it formats all .go files.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -24,10 +24,7 @@ cd $HOME/go/src
 protoc -I=. --go_out=. github.com/googleinterns/step224-2020/config/proto/*.proto
 
 # Run all go tests
-# Find finds all of the files ending in _test.go and returns their relative
-# paths. Sed removes the the file part of that relative path.
-# Then, we remove uniques.
-go test $(find ./github.com/googleinterns/step224-2020 -type f -name '*_test.go' | sed -r 's|/[^/]+$||' -u)
+go test github.com/googleinterns/step224-2020/...
 
 # Run go fmt on all .go files to format them
 go fmt github.com/googleinterns/step224-2020/...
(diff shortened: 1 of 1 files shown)


test.sh
```diff
@@ -27,5 +27,7 @@ protoc -I=. --go_out=. github.com/googleinterns/step224-2020/config/proto/*.prot
 # Find finds all of the files ending in _test.go and returns their relative
 # paths. Sed removes the the file part of that relative path.
 # Then, we remove uniques.
-go test $(find ./github.com/googleinterns/step224-2020
- -type f -name '*_test.go' | sed -r 's|/[^/]+$||' -u)
+go test $(find ./github.com/googleinterns/step224-2020 -type f -name '*_test.go' | sed -r 's|/[^/]+$||' -u)
```
cianbgoog: Alternative thing, that shouldn't be a blocker to merging this:

There must be a better way to deal with this :-P

cianbgoog (approved):

deniadelcid (approved):

test.sh
evanSpendlove: I have added a new version :-P

test.sh
```diff
@@ -24,10 +24,7 @@ cd $HOME/go/src
...
 
 # Run all go tests
-# Find finds all of the files ending in _test.go and returns their relative
-# paths. Sed removes the the file part of that relative path.
-# Then, we remove uniques.
-go test $(find ./github.com/googleinterns/step224-2020 -type f -name '*_test.go' | sed -r 's|/[^/]+$||' -u)
+go test github.com/googleinterns/step224-2020/...
 
 # Run go fmt on all .go files to format them
 go fmt github.com/googleinterns/step224-2020/...
```
deniadelcid: I would swap the order of these two: first run gofmt and then run tests.

deniadelcid (approved):

test.sh
evanSpendlove: That's a good point, thanks, I've reordered these.

Issue #43, opened 2020-08-28
[GCS] Create probe for deleting files in GCS.


evanSpendlove:
- [x] Write .proto file
- [x] Write .pb.go file
- [x] Write .go implementation
- [x] Write _test.go unit test.

evanSpendlove:
Closed by PR #61

PR #51, opened 2020-09-08
Added go implementation of the DeleteFile probe for GCS.

evanSpendlove:
Completes part of #43


gcs/deleteFile/deleteFile.go
```diff
@@ -0,0 +1,90 @@
...
+package gcs
+
+import (
+	"fmt"
+
+	"github.com/google/cloudprober/logger"
+	"github.com/google/cloudprober/metrics"
+	"github.com/google/cloudprober/probes/options"
+
+	deletepb "github.com/googleinterns/step224-2020/gcs/deleteFile/proto"
```
leczb: Just `pb` if this is the only proto that's imported.

gcs/deleteFile/deleteFile.go
```diff
@@ -0,0 +1,90 @@
...
+	// apiCallLatency is used to record latency values with distinct labels
+	// per API call per target.
+	// Recommended usage: append(apiCallLatency[target][apiCall], newMetric)
+	apiCallLatency map[string]map[string][]*metric.EventMetrics
+	logger         *logger.Logger
+
+	bucket string
+	// CancelFunc cancels this probe and its goroutines.
+	// If this probe needs to be cancelled immediately, call CancelFunc().
+	CancelFunc context.CancelFunc
```
leczb: Why is this a field as opposed to a method?

gcs/deleteFile/deleteFile.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// Init initializes the probe with the given parameters.
+// This is a required method to implement the Probe interface specified
```
leczb, line 74: I would include a reference to the interface that it implements: "Implements the <package>.<Type> interface."

gcs/deleteFile/deleteFile.go
```diff
@@ -0,0 +1,90 @@
...
+// by Cloudprober.
+// Arguments:
+//	- name: probe name
+//	- opts: probe configuration options built within Cloudprober
+// Return:
+//	- error: if an error is returned, the config is not a valid delete_file config.
+func (p *Probe) Init(name string, opts *options.Options) error {
+	conf, ok := opts.ProbeConf.(*deletepb.DeleteFileProbe) // Casting ProbeConf interface to DeleteFileProbe type
+	if !ok {
+		return fmt.Errorf("invalid argument: opts.ProbeConf is not of type *deletepb.DeleteFileProbe.")
```
leczb: Nit: No need to punctuate the error message.

gcs/deleteFile/proto/deleteFile.proto
```diff
@@ -12,14 +12,12 @@
...
 //
-// Author: Evan Spendlove (@evanSpendlove)
-//
 // The DeleteFileProbe proto describes the information necessary for a probe to delete a
 // file in GCS.
 
 syntax = "proto2";
 
-package hermes.deletefile;
+package gcs.deletefile;
```
leczb, line 20: It's surprising that the top-level package namespace is not `hermes`.

gcs/deleteFile/proto/deleteFile.proto
```diff
@@ -39,10 +37,6 @@ message DeleteFileProbe {
   // The name of the bucket that contains the object to be deleted.
   optional string bucketName = 2;  
 
-  // REQUIRED
-  // The name of the object to be deleted.
-  optional string objectName = 3;
```
leczb, line 44: It's OK to remove a field that was never used.
When doing removal where it's not guaranteed that the field was never used, mark the removed field name and its tag as reserved to avoid accidentally re-using them:

```
reserved 3;
reserved "objectName";
```

gcs/deleteFile/deleteFile.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/google/cloudprober/logger"
+	"github.com/google/cloudprober/metrics"
+	"github.com/google/cloudprober/probes/options"
+
+	deletepb "github.com/googleinterns/step224-2020/gcs/deleteFile/proto"
+)
+
+// Probe holds aggregate information about all probe runs, per-target.
+// It also holds the config and options used to initialise the probe.
+type Probe struct {
```
leczb: How about calling this DeleteFileProbe?

leczb (changes requested):

gcs/deleteFile/deleteFile.go
```diff
@@ -0,0 +1,90 @@
...
+// TODOs:
+// TODO(evanSpendlove): Setup labels and metrics within probe
+// TODO(evanSpendlove): Add handling of additional labels
+// TODO(evanSpendlove): Add Start() method to Probe.
+// TODO(evanSpendlove): Add initProbeMetrics() method to Probe.
+// TODO(evanSpendlove): Add runProbeForTarget() method to Probe.
+// TODO(evanSpendlove): Add runProbe() method to Probe.
+
+// Package gcs implements the probes required to monitor a GCS instance.
+package gcs
```
deniadelcid, line 28: Not sure whether you've already done so, but just make sure to keep this consistent with the probes @alicjakwie writes.

deniadelcid: Also, having just a bare gcs feels odd (when compared to the rest of your files ), how about `hermes.gcs` or `hermes.probes`?

gcs/deleteFile/deleteFile.go
deniadelcid: Alternatively, if you mean for this to be a generic Probe struct that's applicable to all probes not just deletion probes, then you could rename the file and/or make `config` (below) take in a more generic proto.

gcs/deleteFile/proto/deleteFile.proto
```diff
@@ -39,10 +37,6 @@ message DeleteFileProbe {
   // The name of the bucket that contains the object to be deleted.
   optional string bucketName = 2;
```
deniadelcid, line 38: If this is required (as the comments indicates), shouldn't the label be `required` instead of optional?

gcs/deleteFile/proto/deleteFile.proto
```diff
@@ -39,10 +37,6 @@ message DeleteFileProbe {
   // The name of the bucket that contains the object to be deleted.
   optional string bucketName = 2;  
 
-  // REQUIRED
-  // The name of the object to be deleted.
-  optional string objectName = 3;
-
   // Must extend ProbeDef so this probe can be added to Cloudprober as an
   // extension.
   extend cloudprober.probes.ProbeDef {
```
deniadelcid, line 42: I can't comment on the lines below, but can't you make `DeleteFileProbe` a repeated field instead? This way we can make DeleteFileProbe storage system agnostic (since the `probedef` already carries the info about which storage system we're using). This should remove the need to reserve ranges per storage system. What do you think?

gcs/deleteFile/proto/deleteFile.proto
deniadelcid, line 38: Are "buckets" a GCS specific thing?

deniadelcid (commented):
I was thinking that it doesn't feel "clean" to have `DeleteFileProbe` be a message. Can it instead be an rpc? And the take a [`hermes.HermesProbeDef`](https://github.com/googleinterns/step224-2020/blob/master/config/proto/probe.proto#L30) as part of the request. You could add the `bucketName` as part of `hermes.HermesProbeDef`, if this is something GCS specific, you could rename it to `gcsBucketName`. Finally, you could extend cloudprober with `repeated optional hermes.HermesProbeDef ...`

Otherwise, we would need to make configs for every request type which feels too much work.. Thoughts?

@evanSpendlove @leczb @cianbgoog

gcs/deleteFile/deleteFile.go
evanSpendlove: Ack.

gcs/deleteFile/deleteFile.go
evanSpendlove: This needs to be stored so that it can be called through a method.

I have made this private and added a `CancelProbe()` method.

evanSpendlove: Is it preferable to write the wrapper method rather than exposing this as a public field?

gcs/deleteFile/deleteFile.go
evanSpendlove: Ack.

gcs/deleteFile/proto/deleteFile.proto
evanSpendlove: Okay 😄 Thanks for explaining how to properly remove a proto field when in use!

gcs/deleteFile/deleteFile.go
evanSpendlove: I wasn't sure if you could rename the Probe struct, but I looked into it a bit further and it looks like you can without breaking Cloudprober compatibility.

evanSpendlove: I've renamed this to `DeleteFileProbe`.

gcs/deleteFile/deleteFile.go
evanSpendlove: I'm not quite sure what you mean by including a reference to the interface.

There is a `Probe` interface specified in probes/probes.go in Cloudprober.

https://github.com/google/cloudprober/blob/baf63376da5f8b54705ac723f58c5233d8dd68cf/probes/probes.go#L67

This is the interface that is implemented.

Are you suggesting that I add a link to reference this, or a file path?

gcs/deleteFile/proto/deleteFile.proto
evanSpendlove: It is required, but the recommendation is to avoid the use of `required` in proto2 as per the style guide.

evanSpendlove: "Bucket" is the term used in GCS, but it is not GCS specific, rather object-storage specific as object storage systems use the concept of "buckets" and "objects". For example, both GCS and Ceph use buckets and objects.

gcs/deleteFile/deleteFile.go
cianbgoog: I think you should include a line in the comment on the DeleteFileProbe struct to indicate that it implements the Cloudprober.Probes.Probe interface.

evanSpendlove:
We are no longer pursuing individual probes. We have moved to a single probe approach.

PR #52, opened 2020-09-09
Restructured packages and removes DeleteFileProbe proto.

evanSpendlove:
Alicja and I have discussed the package structure for Hermes and confirmed our approach.

This PR restructures the existing packages, and removes the DeleteFileProbe proto.

We are moving towards a single proto message which is used in each probe implementation, so we no longer need a new proto per probe.


client/client.go
```diff
@@ -19,7 +19,8 @@
 // TODO(#38): Add support for secure connection to RPC server with credentials, if supported by Cloudprober.
 // TODO(evanSpendlove): Include details of these errors (addProbeFromConfig method) in the error message.
 
-package cmd
+// Client contains the gRPC client implementation used to send RPCs to Cloudprober.
+package client
```
cianbgoog, line 25: If you're changing package names, you should change their path as well.

hermes/shared/file_gen.go
```diff
@@ -7,11 +7,9 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-
-// Author: Alicja Kwiecinska (kwiecinskaa@google.com) github: alicjakwie
-
+//
 // File generation for Hermes
-package hermes
+package shared
```
cianbgoog: Change the file path as well.

cianbgoog (changes requested):

client/client.go
evanSpendlove: The path for this file has been changed. It is now in the directory `client`, so its path is: `step224-2020/client/client.go`.

hermes/shared/file_gen.go
evanSpendlove: The path for this file has been changed. It is now in the directory `shared`, so its path is: `step224-2020/hermes/shared/file_gen.go`.

leczb: "shared" has a code-smell (it's similar to "misc" and "util"). What's wrong with "hermes"?

hermes/shared/file_gen.go
```diff
@@ -7,11 +7,9 @@
...
 
 import (
 	"crypto/sha1"
```
leczb: This comment applies to lines 24 and below:

The documentation comments above functions/methods should start with the function name (exactly).

hermes/shared/file_gen.go
evanSpendlove: We can move it to the "hermes" package. I'll adjust the file path accordingly.

hermes/file_gen.go
```diff
@@ -7,11 +7,9 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-
-// Author: Alicja Kwiecinska (kwiecinskaa@google.com) github: alicjakwie
-
+//
 // File generation for Hermes
```
leczb: Package comments should start with "Package <package name> ...".

hermes/file_gen.go
```diff
@@ -7,11 +7,9 @@
...
-package hermes
+package shared
 
 import (
 	"crypto/sha1"
```
leczb, line 23: This is for line 19: exported identifiers should all have doc comments.

leczb (changes requested):

hermes/shared/file_gen.go
evanSpendlove: I've amended these comments to these specifications.

hermes/file_gen.go
evanSpendlove: Ack.

hermes/file_gen.go
evanSpendlove: Ack.

hermes/file_gen.go
```diff
@@ -1,51 +1,70 @@
...
+	fileContents := []byte(f.contents)
+	hash := sha1.Sum(fileContents)
 	// return checksum in hex notation
 	return fmt.Sprintf("%x", hash)
 }
 
-// method of HermesFile generates the file takes id as a parameter
+// GenerateHermesFile generates a HermesFile object with a valid file name and contents.
+// Arguments:
+//	- id: id must be >= 0 and <= 49.
```
leczb, line 62: This should be checked in code.

hermes/file_gen.go
```diff
@@ -1,51 +1,70 @@
...
+// Returns:
+//	- HermesFile: returns a HermesFile object with a valid file name and contents.
 func GenerateHermesFile(id int) HermesFile {
```
leczb, line 65: Nit: The idiomatic name for this would be NewHermesFile(). Even better: move the HermesFile related code to a `file` (or `hermesfile`) package and call this New(). Then the call-site will look like this:

`f := hermesfile.New(id)`

hermes/file_gen.go
```diff
@@ -1,51 +1,70 @@
...
 	file := HermesFile{}
 	file.generateFileContents(id)
```
leczb, line 67: Not that this is wrong, but I think it's cleaner to have a stand-alone function that generates the contents, then this would be 

`file.contents = generateContents(id)`

...which reads better. The current version hides what's actually happening.

hermes/file_gen_test.go
```diff
@@ -25,18 +28,18 @@ type FileNameTestUnit struct {
 	want     string
 }
 
-var test_unit_table [5]FileNameTestUnit
+var testUnitTable [5]FileNameTestUnit
 
-// Function completes the table of HermesFileTestUnit by iterating over its entries
+// completeTestUnitTable completes the table of HermesFileTestUnit by iterating over its entries
```
leczb, line 33: Avoid logic in test code. This should be a static slice instead, declared within the test function that uses it.
In general: move the declarations closest to their use.

hermes/file_gen_test.go
```diff
@@ -25,18 +28,18 @@ type FileNameTestUnit struct {
...
-		test_unit_table[i].checksum = "abba"
-		test_unit_table[i].want = fmt.Sprintf("Hermes_%02d_%v", test_unit_table[i].id, test_unit_table[i].checksum)
+		testUnitTable[i].id = i*10 + 1
+		testUnitTable[i].checksum = "abba"
+		testUnitTable[i].want = fmt.Sprintf("Hermes_%02d_%v", testUnitTable[i].id, testUnitTable[i].checksum)
 	}
 }
 
-// Test function for the generateFileChecksum function in file_gen.go
+// TestChecksum tests the generateFileChecksum() method.
```
leczb, line 42: This is for line 49:
The idiomatic error message would be

`t.Errorf("generateFileChecksum() = %v; want %v")`

Also note that unit tests should exercise the public API of the package. In rare cases it makes sense to test some non-exported function/method. In this case it's not necessary.

hermes/file_gen_test.go
```diff
@@ -47,15 +50,15 @@ func TestChecksum(t *testing.T) {
 	}
 }
 
-// Test function for the generateFileName function in file_gen.go
+// TestFileName tests the generateFileName method.
```
leczb, line 53: It's not strictly required to comment exported functions in _test.go files.
Only add comments where the name is not self-describing or something unusual/complicated is going on.

leczb (changes requested):

hermes/file_gen.go
alicjakwie: I had this in a previous version but during a review was told that it would be better to make generateFileContents and generateFileName methods of the struct HermesFile. What is in your opinion more readable?

hermes/file_gen.go
alicjakwie: Thank you will add that!

hermes/file_gen.go
leczb: The name "generateFileContents" suggests that its output is the contents.
However, the current version also sets the contents on the file, which is surprising.

Methods make sense when they access one or more fields of the receiver as an input or they use a mutex in the receiver.
generateFileContents() doesn't really use the file as an input.

client/client.go
```diff
@@ -12,14 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+// Author: Evan Spendlove, espendlove@google.com, GitHub: evanSpendlove.
```
deniadelcid: Just make sure that you want to make your emails public before merging. It's totally up to you.

hermes/file_gen.go
alicjakwie: Okay thank you! will go back to the previous version then.

hermes/file_gen_test.go
alicjakwie: For now the file contents are just a simple string but I will change it soon and we will use read.io and a pseudo random byte generator, therefore I wanted to keep the TestChecksum function in the test file. Is this alright?

hermes/file_gen_test.go
alicjakwie: Would it be better to not export them then? Will delete unnecessary comments!

hermes/file_gen_test.go
leczb: I would implement a single test that creates HermesFile instances and checks that the filename, contents and checksum have the expected values.
The tests should exercise the system to the fullest extent feasible. If the test only exercise individual functions, then there's no guarantee that the package-level behaviour is correct.
For example, this test passes, but then someone deletes the line that calls this function. The result will be an empty checksum and no test failures.

deniadelcid (approved):

hermes/file_gen_test.go
deniadelcid: In this case, you need to export them since these are your test functions. If you don't then all your tests will pass, but none will actually run. Just remove the comments.

hermes/file_gen_test.go
alicjakwie: I would like to keep these test functions to test the individual components here on their own since the rest of Hermes's logic depends on the correctness of these components but will add the function that tests everything together once the read.io integration is complete!

hermes/file_gen.go
evanSpendlove: @alicjakwie is making these changes in a different PR: https://github.com/googleinterns/step224-2020/pull/58 so I'll mark this resolved for now.

hermes/file_gen.go
evanSpendlove: @alicjakwie is making these changes in a different PR: https://github.com/googleinterns/step224-2020/pull/58 so I'll mark this resolved for now.

hermes/file_gen_test.go
evanSpendlove: @alicjakwie is making these changes in a different PR: https://github.com/googleinterns/step224-2020/pull/58 so I'll mark this resolved for now.

hermes/file_gen_test.go
evanSpendlove: @alicjakwie is making these changes in a different PR: https://github.com/googleinterns/step224-2020/pull/58 so I'll mark this resolved for now.

hermes/file_gen_test.go
evanSpendlove: @alicjakwie is making these changes in a different PR: https://github.com/googleinterns/step224-2020/pull/58 so I'll mark this resolved for now.

client/client.go
evanSpendlove: I've thought this over and removed it. Thank you.

hermes/file_gen.go
evanSpendlove: @alicjakwie is making these changes in a different PR: #58 so I'll mark this resolved for now.

config/proto/probe.proto
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+// Author: Evan Spendlove, GitHub: @evanSpendlove.
```
cianbgoog, line 15: Do we need author comments? Here and elsewhere.

hermes/file_gen_test.go
```diff
@@ -25,18 +28,18 @@ type FileNameTestUnit struct {
 	want     string
 }
 
-var test_unit_table [5]FileNameTestUnit
+var testUnitTable [5]FileNameTestUnit
```
cianbgoog, line 31: nit: It's somewhat more normal to define these kind of structs in place (because they often require different things for want & input)

cianbgoog (approved):
Just a few nits - happy for you to merge once you've responded.

leczb (approved):

config/proto/probe.proto
evanSpendlove: Alicja and I have discussed this with Balazs, and we're happy to go with author comments in all files.

hermes/file_gen_test.go
evanSpendlove: Alicja is handling comments on file_gen.go and file_gen_test.go in a different PR, so I'll add this comment there.

evanSpendlove: Ah, it's already there!

PR #54, opened 2020-09-10
Modified probe.proto and targets.proto to allow for a single unified probe proto.

evanSpendlove:


config/proto/probe.proto
```diff
@@ -39,28 +38,19 @@ message HermesProbeDef {
...
-    PUT = 2;  // PUT probe is for putting an object (file) into a storage system.
-    DELETE = 3;  // DELETE probe is for deleting an object (file) from a storage system.
-  }
-
-  TargetSystem target_system = 3;
-  ProbeType probe_type = 4;
+  optional TargetSystem target_system = 3;
 
-  int32 interval_sec = 5;  // Probing interval in seconds, default = 3600
+  optional int32 interval_sec = 4;  // Probing interval in seconds, default = 3600
```
leczb: Nit: Move the comments above the fields.

config/proto/probe.proto
```diff
@@ -72,10 +62,15 @@ message HermesProbeDef {
     HOUR = 6;
   }
   
-  LatencyUnit latency_unit = 9;  // default is MICROSECOND
+  optional LatencyUnit latency_unit = 8;  // default is MICROSECOND
```
leczb: Can you explain the purpose of this field?
I expect the cloudprober.metrics.Dist fields hold float values. Using seconds as the unit is the natural choice.

config/proto/targets.proto
```diff
@@ -49,4 +49,5 @@ message Target {
 
   int64 total_space_allocated_mib = 5;
   string target_url = 6;  // URL for connecting to the the API of the target storage system.
+  string hermes_bucket_name = 7;  // Name for bucket used by Hermes on this target storage system.
```
leczb: I don't think the "hermes" prefix is needed here.

leczb (changes requested):

config/proto/probe.proto
evanSpendlove: The purpose of this field is to allow the user to define what latency unit to use for the latency metrics.
Yes, cloudprober.metrics.Dist fields do hold float64 values.

Seconds is a natural choice, but I did some testing on GCE, and the latency of an equivalent to ListBuckets() was 2 microseconds, which was a lot more representative to me as 2us, than 2e-6.

I'm open to changing this though, if you think that seconds would be more suitable, I can change it to that.

leczb: 2us is a human-readable rendering of the 2e-6 value.
In general: only convert something to human-readable form at the very last step (for example when rendering an HTML template or printing something in a CLI). The values should be transmitted/persisted in their rawest form.

config/proto/probe.proto
```diff
@@ -39,28 +38,19 @@ message HermesProbeDef {
     CEPH = 2;
```
deniadelcid: Should we remove this line (at least for now) given that support for Ceph is on pause?

deniadelcid (approved):

config/proto/probe.proto
evanSpendlove: Okay, that's a very good point. Thank you. I'll stick with seconds until the last step of surfacing metrics.

evanSpendlove: I think this field is still important though, as it decides the unit of the latency metric (which is included in the metric name).

However, I will change the default unit to seconds.

config/proto/targets.proto
evanSpendlove: Ack.

config/proto/probe.proto
evanSpendlove: Yes, ack.

config/proto/probe.proto
evanSpendlove: Ack.

config/proto/targets.proto
```diff
@@ -27,13 +27,14 @@ message Target {
   // TargetSystem is expected to be a storage system.
   // It will stay constant until support for new storage systems is added.
   enum TargetSystem {
-    GOOGLE_CLOUD_STORAGE = 0;
-    CEPH = 1;
+    TARGET_SYSTEM_UNSPECIFIED = 0;
```
cianbgoog, line 32: Could you just write `UNKNOWN` here, and elsewhere.

cianbgoog (approved):
minor nits.

config/proto/probe.proto
leczb: What do you mean by "decides the unit of the latency metric"? Is this a Cloudprober thing?

leczb (approved):

config/proto/probe.proto
evanSpendlove: There are two latency distribution metrics: apiLatency and probeLatency. These both stores latency of operations.These will be metrics exported to Prometheus. As such, they need a latency unit. Cloudprober allows the user to configure the unit for the latency metrics (s, ms, us, ns, etc.). We decided to also support this feature.

config/proto/targets.proto
evanSpendlove: I think the convention in protobuffers is to use the enum name with the suffix "UNSPECIFIED" for the 0 value. This is handled nicely in the pb.go file I think when [using the generated getters](https://github.com/googleinterns/step224-2020/blob/6fa6e5ecb5aa224830efa625c075943ba03fed9c/config/proto/probe.pb.go#L293-L298).

Convention source: [Enum in Proto](https://developers.google.com/protocol-buffers/docs/style#enums)

config/proto/probe.proto
evanSpendlove: I've removed the latency unit from the proto and added the measurement unit in the comments for the relevant proto fields.

config/proto/probe.proto
```diff
@@ -44,26 +44,15 @@ message HermesProbeDef {
   // Probes will not retry, they will fail if they timeout.
   optional int32 timeout_sec = 5;  // Timeout in seconds, default = 60
 
-  // If specified, latency is stored as a distribution
+  // If specified, latency is stored as a distribution metric.
```
leczb, line 49: Note the measurement unit in the comments of distribution metric fields.

alicjakwie (approved):

config/proto/probe.proto
evanSpendlove: Ah I do, but it's just slightly lower down 😛 in the code. 

"// The measurement unit of latency will be seconds."
on lines 49 and 53.

PR #59, opened 2020-09-10
Added struct values and Init() to monitor.go

evanSpendlove:


hermes/probe/monitor.go
```diff
@@ -12,9 +12,51 @@
...
+}
+
+// Init initializes the probe with the given parameters.
+// This is a required method to implement the cloudprober.Probes.Probe interface.
+func (mp *MonitorProbe) Init(name string, opts *options.Options) error {
+	conf, ok := opts.ProbeConf.(*pb.HermesProbeDef) // Casting ProbeConf interface to HermesProbeDef type
+	if !ok {
+		return fmt.Errorf("invalid argument: opts.ProbeConf is not of type *pb.HermesProbeDef")
+	}
+	p.config = conf
```
deniadelcid: I think in here and the rest of the lines below, you meant `mp`. `p` doesn't appear to be declared.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,51 @@
...
+)
+
 // MonitorProbe holds aggregate information about all probe runs, per-target.
 // It also holds the config and options used to initialise the probe.
-type MonitorProbe struct{}
+type MonitorProbe struct {
+	name    string
+	config  *pb.HermesProbeDef
+	targets []string
+	opts    *options.Options
```
deniadelcid: not clear where does `options` come from in here. I think you should include its path in your imports.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,51 @@
...
+
+	// probeOpLatency is used to record latency values with distinct labels
+	// per probe operation per target.
+	// Recommended usage: append(probeOpLatency[target][probeOp], newMetric)
+	probeOpLatency map[string]map[string][]*metrics.EventMetrics
```
deniadelcid: `metrics` is also not imported.

deniadelcid (commented):
- [X] Can you run gofmt? It seems like some import statements are missing and there are minor mismatches between your variable names, gofmt should catch these.

- [X] Is there already a test file for this? If not, could you start one?

-- REPLY -- 
For some reason, `go fmt` ignored these? But I've fixed the imports and variable mismatch anyways.

I've started a test file, but I believe @alicjakwie will be writing the tests for `Init()` and `Start()`.

hermes/probe/monitor.go
alicjakwie: Could you add a comment of what targets []string are? since there's also config.targets this could be confusing for the user (at least it was for me as you know)

hermes/probe/monitor.go
alicjakwie:
```suggestion
	mp.config = conf
```

hermes/probe/monitor.go
```diff
@@ -12,9 +12,51 @@
...
+
+// Init initializes the probe with the given parameters.
+// This is a required method to implement the cloudprober.Probes.Probe interface.
+func (mp *MonitorProbe) Init(name string, opts *options.Options) error {
+	conf, ok := opts.ProbeConf.(*pb.HermesProbeDef) // Casting ProbeConf interface to HermesProbeDef type
+	if !ok {
+		return fmt.Errorf("invalid argument: opts.ProbeConf is not of type *pb.HermesProbeDef")
+	}
+	p.config = conf
+	p.name = name
```
alicjakwie:
```suggestion
	mp.name = name
```

alicjakwie:
```suggestion
	mp.name = name
```

hermes/probe/monitor.go
```diff
@@ -12,9 +12,51 @@
...
+	p.opts = opts
```
alicjakwie:
```suggestion
	mp.opts = opts
```

hermes/probe/monitor.go
```diff
@@ -12,9 +12,51 @@
...
+	p.probeOpLatency = make(map[string]map[string]*metrics.EventMetrics)
```
alicjakwie:
```suggestion
	mp.probeOpLatency = make(map[string]map[string]*metrics.EventMetrics)
```

hermes/probe/monitor.go
```diff
@@ -12,9 +12,51 @@
...
+	p.apiCallLatency = make(map[string]map[string]*metrics.EventMetrics)
```
alicjakwie:
```suggestion
	mp.apiCallLatency = make(map[string]map[string]*metrics.EventMetrics)
```

hermes/probe/monitor.go
```diff
@@ -12,9 +12,51 @@
...
+	p.logger = opts.Logger
```
alicjakwie:
```suggestion
	mp.logger = opts.Logger
```

hermes/probe/monitor.go
```diff
@@ -12,9 +12,51 @@
...
+// - Alicja Kwiecinska (kwiecinskaa@google.com) github: alicjakwie
+// - Evan Spendlove, GitHub: evanSpendlove.
+//
 // Monitor implements a probe that monitors a storage system using the
 // Hermes algorithm.
 
+package probe
+
+import (
+	pb "github.com/googleinterns/step224-2020/config/proto"
```
alicjakwie: Shouldn't options and metrics be imported as well?

hermes/probe/monitor.go
evanSpendlove: Yes, sorry, refactoring doesn't always work the way I want it to in VScode.

evanSpendlove: I've amended this in the latest commit.

hermes/probe/monitor.go
evanSpendlove: I've amended my imports in my latest commit.

hermes/probe/monitor.go
evanSpendlove: I've amended my imports in my latest commit.

Also, I've changes targets to an `[]*pb.Target` as I feel that this will be more useful.

hermes/probe/monitor.go
evanSpendlove: Ack.

hermes/probe/monitor.go
evanSpendlove: Ack.

hermes/probe/monitor.go
evanSpendlove: Ack.

hermes/probe/monitor.go
evanSpendlove: Ack.

hermes/probe/monitor.go
evanSpendlove: Ack.

hermes/probe/monitor.go
evanSpendlove: Ack.

hermes/probe/monitor.go
evanSpendlove: Ack.

hermes/probe/monitor.go
evanSpendlove: You can comment on multiple lines by click the + on the first one and then holding SHIFT and clicking the + on the last one.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,57 @@
...
+func (mp *MonitorProbe) Init(name string, opts *options.Options) error {
+	conf, ok := opts.ProbeConf.(*pb.HermesProbeDef) // Casting ProbeConf interface to HermesProbeDef type
+	if !ok {
+		return fmt.Errorf("invalid argument: opts.ProbeConf is not of type *pb.HermesProbeDef")
+	}
+	mp.config = conf
+	mp.name = name
+	mp.opts = opts
+	mp.probeOpLatency = make(map[string]map[string]*metrics.EventMetrics)
+	mp.apiCallLatency = make(map[string]map[string]*metrics.EventMetrics)
```
cianbgoog: If you know the list of metric labels you'll have, then it's considered good practise in prometheus to initialise them to empty metrics.

cianbgoog (approved):

hermes/probe/monitor.go
```diff
@@ -12,9 +12,139 @@
...
+	apiCallLatency map[string]map[string][]*metrics.EventMetrics
+	logger         *logger.Logger
+}
+
+// Init initializes the probe with the given parameters.
+// This is a required method to implement the cloudprober.Probes.Probe interface.
+func (mp *MonitorProbe) Init(name string, opts *options.Options) error {
+	conf, ok := opts.ProbeConf.(*probepb.HermesProbeDef) // Casting ProbeConf interface to HermesProbeDef type
+	if !ok {
+		return fmt.Errorf("invalid argument: opts.ProbeConf is not of type *pb.HermesProbeDef")
```
evanSpendlove: Should be `of type *probepb.HermesProbeDef`

hermes/probe/monitor.go
```diff
@@ -12,9 +12,139 @@
...
+			for _, target := range mp.apiCallLatency {
+				for _, call := range target {
+					for _, metric := range call {
+						metricChan <- metric
+					}
+				}
+			}
+
+			probeCtx, _ := context.WithDeadline(ctx, time.Now().Add(mp.opts.Timeout))
+			mp.initProbeMetrics() // init during probe run as these are timestamped.
```
evanSpendlove: What is "these"? Should be "metrics".

hermes/probe/monitor.go
```diff
@@ -12,9 +12,139 @@
...
+			mp.runProbe(probeCtx)
+		}
+	}
+}
+
+// TODO: Add comment here explaining function
+func (mp *MonitorProbe) initLatencyDistMetric(target string, latencyMap map[string]map[string][]*metrics.EventMetrics, latDistPb *metricpb.Dist, operations []string) {
+	latDist, err := metrics.NewDistributionFromProto(latDistPb)
```
evanSpendlove: Could add if clause here and use break for checking if nil.

hermes/probe/monitor_test.go
```diff
@@ -0,0 +1,118 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Authors:
+// - Alicja Kwiecinska (kwiecinskaa@google.com) github: alicjakwie
+// - Evan Spendlove, GitHub: evanSpendlove.
+//
+// Monitor_test tests the functions used in the monitor probe, as well
+// as the probes integration with and use within Cloudprober.
```
evanSpendlove:
```suggestion
// as the probes integration with Cloudprober.
```

hermes/probe/monitor_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+package probe
```
evanSpendlove: Add Package comment.

hermes/probe/monitor_test.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+// genTestProbeConf generates a test HermesProbeDef proto config for
+// initialising and running a MonitorProbe.
+// Arguments:
+//	- name: pass the name for this probe instance.
+// Returns:
+//	- probeDef: a test probe config with a HermesProbeDef extension set.
+func genTestConfig(name string) (*probes_configpb.ProbeDef, *monitorpb.HermesProbeDef) {
+	probeDef := &probes_configpb.ProbeDef{ // Create probe def
```
evanSpendlove: Remove unnecessary comment.

hermes/probe/monitor_test.go
```diff
@@ -0,0 +1,118 @@
...
+		Name: proto.String(name),
+		Type: probes_configpb.ProbeDef_EXTENSION.Enum(),
+	}
+
+	// Add HermesProbeDef extension to probeDef
```
evanSpendlove: Remove unnecessary comment.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,139 @@
...
+func (mp *MonitorProbe) Start(ctx context.Context, metricChan chan *metrics.EventMetrics) {
+	probeTicker := time.NewTicker(mp.opts.Interval)
+
+	for {
+		select {
+		case <-ctx.Done():
+			probeTicker.Stop()
+			return
+		case <-probeTicker.C:
+			// On probe tick, write data to the channel and run probe.
```
evanSpendlove: Remove unnecessary comment.

hermes/probe/monitor.go
evanSpendlove: Ack.

hermes/probe/monitor.go
evanSpendlove: Ack.

hermes/probe/monitor.go
evanSpendlove: Ack.

hermes/probe/monitor.go
evanSpendlove: Ack, change made.

hermes/probe/monitor_test.go
evanSpendlove: Ack.

evanSpendlove (commented):
Added minor comments.

hermes/probe/monitor_test.go
evanSpendlove: Ack.

hermes/probe/monitor_test.go
evanSpendlove: Ack.

hermes/probe/monitor_test.go
evanSpendlove: Ack.

hermes/probe/monitor.go
evanSpendlove: I'm not quite sure what you mean by this. Do you mean to add the labels without values for each metric?

Metric labels are initialised in the `initProbeMetrics()` function.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,155 @@
...
 // limitations under the License.
 //
+// Authors:
+// - Alicja Kwiecinska (kwiecinskaa@google.com) github: alicjakwie
+// - Evan Spendlove, GitHub: evanSpendlove.
+//
 // Monitor implements a probe that monitors a storage system using the
 // Hermes algorithm.
 
+// Package probe implements the probe that Hermes uses to monitor
```
leczb: Nit: no need to break the line.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,155 @@
...
+	// per API call per target.
+	// Recommended usage: append(apiCallLatency[target][apiCall], newMetric)
+	apiCallLatency map[string]map[string][]*metrics.EventMetrics
+	logger         *logger.Logger
+}
+
+// Init initializes the probe with the given parameters.
+// This is a required method to implement the cloudprober.Probes.Probe interface.
+func (mp *MonitorProbe) Init(name string, opts *options.Options) error {
+	conf, ok := opts.ProbeConf.(*probepb.HermesProbeDef) // Casting ProbeConf interface to HermesProbeDef type
```
leczb: Nit: There's no need for this comment, as it just repeats what the code does.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,155 @@
...
+	mp.logger = opts.Logger
+
+	return nil
+}
+
+// Start runs the probe indefinitely, unless cancelled, at the configured interval.
+// Probe metrics will be sent via the metricChan at the beginning of the next probe run.
+// Arguments:
+//	- ctx: context provided for cancelling probe.
+//	- metricChan: bidirectional channel used for sending metrics to be surfaced.
```
leczb: Why is it bidirectional?

hermes/probe/monitor.go
```diff
@@ -12,9 +12,155 @@
...
+func (mp *MonitorProbe) Start(ctx context.Context, metricChan chan *metrics.EventMetrics) {
+	probeTicker := time.NewTicker(mp.opts.Interval)
+
+	for {
+		select {
+		case <-ctx.Done():
+			probeTicker.Stop()
+			return
+		case <-probeTicker.C:
+			for _, target := range mp.probeOpLatency {
```
leczb: This function is too long. I would but these two for loops in a separate function (reportMetrics() maybe).

hermes/probe/monitor.go
```diff
@@ -12,9 +12,155 @@
...
+					for _, metric := range call {
+						metricChan <- metric
+					}
+				}
+			}
+
+			probeCtx, _ := context.WithDeadline(ctx, time.Now().Add(mp.opts.Timeout))
+			// init during probe run as metrics are timestamped.
+			if err := mp.initProbeMetrics(); err != nil {
+				mp.logger.Errorf("error initialising probe metric: %v", err)
```
leczb: Is it OK to go ahead with running runProbe() if initProbeMetrics() failed?

hermes/probe/monitor.go
```diff
@@ -12,9 +12,155 @@
...
+func (mp *MonitorProbe) Init(name string, opts *options.Options) error {
+	conf, ok := opts.ProbeConf.(*probepb.HermesProbeDef) // Casting ProbeConf interface to HermesProbeDef type
+	if !ok {
+		return fmt.Errorf("invalid argument: opts.ProbeConf is not of type *probepb.HermesProbeDef")
+	}
+	mp.name = name
+	mp.config = conf
+	mp.targets = mp.config.GetTargets()
+	mp.opts = opts
+	mp.probeOpLatency = make(map[string]map[string][]*metrics.EventMetrics)
```
leczb: Is it guaranteed that there's no concurrent access to these maps?
If there is concurrent access, then add a mutex.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,155 @@
...
+			mp.runProbe(probeCtx)
+		}
+	}
+}
+
+// initLatencyDistMetric initialises a latency distribution metric
+// according to the values passed.
+// Arguments:
+//	- target: pass target name.
+//	- latencyMap: pass a map with the key format: map[target][operation].
```
leczb: Make it clear that this is the output.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,155 @@
...
+			AddLabel("storage_system", mp.config.GetTargetSystem().String()).
+			AddLabel("probe_operation_type", op).
+			AddLabel("target", target))
+	}
+	return nil
+}
+
+// initProbeMetrics initializes missing latency metrics for each target.
+func (mp *MonitorProbe) initProbeMetrics() error {
+	for _, target := range mp.targets {
```
leczb: Nit: Use short variable names for small-scoped variables. Just `t` would do here instead of `target`.

hermes/probe/monitor_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	opts.ProbeConf = cfg
+
+	return opts
+}
+
+func TestInit(t *testing.T) {
+	mp := &MonitorProbe{}
+	_, hermesConf := genTestConfig("testProbe")
+	mp.Init("testProbe", genOptsFromConfig(t, hermesConf))
```
leczb: Add some checks.
At the minimum, check the error returned by Init().

hermes/probe/monitor_test.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func TestStart(t *testing.T) {
```
leczb: Remove empty tests.

leczb (changes requested):

hermes/probe/monitor.go
evanSpendlove: Okay, I left the comment there as Denia had previously asked what the code does and it's not the most intuitive line. I'll remove it.

hermes/probe/monitor.go
evanSpendlove: Okay, thank you. I'll do that.

hermes/probe/monitor.go
evanSpendlove: There could be concurrent access to these maps. However, no two goroutines would try to access the same data point, as the keys for the first map are the targets, and goroutines are spawned per target.

Is a mutex still needed here?

hermes/probe/monitor.go
evanSpendlove: Okay, thank you. I'll amend this.

hermes/probe/monitor.go
evanSpendlove: The channel passed by Cloudprober is bidirectional, but I realise now that we can define the channel arg to be unidirectional, and it will still work. We won't ever need to receive data via the channel.

hermes/probe/monitor.go
evanSpendlove: Ack.

hermes/probe/monitor.go
leczb: If the concurrent access is read-only, it's OK.
If the concurrent access involves update/delete, a mutex is required. 
See https://golang.org/doc/faq#atomic_maps

hermes/probe/monitor.go
leczb: I believe she had a question here as it's not obvious why you would do this cast.
Casts have a code-smell. Any time they are required, we should add a comment explaining the rationale.
Repeating *what* the code does in the comment makes the comment redundant. The comment should explain the *why*.

hermes/probe/monitor.go
evanSpendlove: Ack.

hermes/probe/monitor.go
evanSpendlove: Unfortunately, we cannot return the error as the Start() method defined in the Probe interface this implemented does not return an error. Also, Cloudprober does not provide error handling when calling this function.

leczb: What I mean was that line 111 is executed even if an error happens here.
Is that the intention?

hermes/probe/monitor_test.go
evanSpendlove: Removed.

hermes/probe/monitor_test.go
evanSpendlove: I will add full checks to this test.

evanSpendlove: I will add full checks to this test.

hermes/probe/monitor.go
evanSpendlove: Okay, I will add a mutex. Thank you.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,162 @@
...
+	mp.apiCallLatency = make(map[string]map[string][]*metrics.EventMetrics)
+	mp.logger = opts.Logger
+
+	return nil
+}
+
+// reportMetrics sends the metrics recorded in the last probe run to Cloudprober.
+// Arguments:
+//	- metricChan: metric channel passed from Cloudprober
+func (mp *MonitorProbe) reportMetrics(metricChan chan<- *metrics.EventMetrics) {
```
cianbgoog: Do you need to do anything when you call reportMetrics to stop yourself from reporting the same metrics again the next time it's called?

hermes/probe/monitor.go
```diff
@@ -12,9 +12,162 @@
...
+func (mp *MonitorProbe) Start(ctx context.Context, metricChan chan<- *metrics.EventMetrics) {
+	probeTicker := time.NewTicker(mp.opts.Interval)
+
+	for {
+		select {
+		case <-ctx.Done():
+			probeTicker.Stop()
+			return
+		case <-probeTicker.C:
+			mp.reportMetrics(metricChan)
```
cianbgoog: Could you do this the opposite way around, and return the metrics from mp.runProbe(), then immediately report them?

cianbgoog (changes requested):

hermes/probe/monitor.go
leczb: General note: make sure the code compiles (and unit tests pass) before sending the change for review.

hermes/probe/monitor.go
leczb: +1
It would be better to only report metrics after a successful probe iteration. It's easier to do that in runProbe().
Either there, or do it right after runProbe().

hermes/probe/monitor.go
leczb: I don't see any change.
Please only mark a conversation resolved when the requested change was made.

evanSpendlove: I did make a change to this in a commit on Friday:

I added: "-> This is the latency map being initialised with metrics and labels." beneath the argument comment.

However, I have made significant changes to the structure of metrics and removed this function since then.

hermes/probe/monitor.go
evanSpendlove: I think I understand what you mean, and this is done when initialising the metrics.

hermes/probe/monitor.go
evanSpendlove: I understand what you mean. I have added a `return` here as we cannot proceed with running the probe if the probe does not initialise the metrics properly.

hermes/probe/monitor.go
evanSpendlove: I have refactored the metrics and included a mutex.

hermes/probe/monitor.go
evanSpendlove: I have refactored the metrics which removes this issue, but thank you for pointing this out!

hermes/probe/monitor.go
evanSpendlove: I've moved metric reporting into runProbe() as suggested.

hermes/probe/metric.go
```diff
@@ -0,0 +1,107 @@
...
+	// per exit status per probe operation per target.
+	// Recommended usage: probeOpLatency[probeOp][exitStatus].Metric("latency").AddFloat64(<val>)
+	ProbeOpLatency map[string]map[string]*metrics.EventMetrics
+	// apiCallLatency is used to record latency values with distinct labels
+	// per exit status per API call per target.
+	// Recommended usage: apiCallLatency[apiCall][exitStatus].Metric("latency").AddFloat64(<val>)
+	ApiCallLatency map[string]map[string]*metrics.EventMetrics
+}
+
+// NewProbeMetrics creates a new *probeRunMetrics object and initialises the fields inside it.
```
leczb: The convention is to match the name of the New* function with the object it returns (or the name of the interface the object implements).
Also, don't mention unexported symbols in comments, as those are included in the generated GoDoc.

hermes/probe/metric.go
```diff
@@ -0,0 +1,107 @@
...
+// Arguments:
+//	- conf: pass a HermesProbeDef config
+//	- target: pass the target for which metrics are to be collected.
+// Returns:
+//	- pm: returns an initialised *probeRunMetrics object.
+//	- err: returns an error if a latency distribution cannot be created from the config proto.
+func NewProbeMetrics(conf probepb.HermesProbeDef, target probepb.Target) (*probeRunMetrics, error) {
+	probeOps := []string{"total_probe_run", "check_nil", "read_and_verify", "delete_file", "create_file"}
```
leczb: If these strings are going to be used elsewhere, put them in consts.
Nit: put each value on a separate line.

hermes/probe/metric.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/google/cloudprober/metrics"
+	"time"
+
+	metricpb "github.com/google/cloudprober/metrics/proto"
+	probepb "github.com/googleinterns/step224-2020/config/proto"
+)
+
+// probeRunMetrics stores the cumulative metrics for probe runs for a target.
+// These metrics are sent via a data channel at the end of the probe run.
+type probeRunMetrics struct {
```
leczb: Does this have be distinguished from other metrics? If not I would simplify the name to "Metrics".

hermes/probe/metric.go
```diff
@@ -0,0 +1,107 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Author: Evan Spendlove, GitHub: evanSpendlove.
+//
+// Metric implements the latency distribution metrics for Hermes.
+// Package probe implements the probe that Hermes uses to monitor
+// a storage system.
+
+package probe
```
leczb: Consider moving the metrics to a separate package.

hermes/probe/metric.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"fmt"
+	"github.com/google/cloudprober/metrics"
```
leczb: Nit: put this import into a separate group.
Import should be grouped like this:
1) standard library packages
2) other packages
3) renamed packages

hermes/probe/metric.go
```diff
@@ -0,0 +1,107 @@
...
+	"time"
+
+	metricpb "github.com/google/cloudprober/metrics/proto"
+	probepb "github.com/googleinterns/step224-2020/config/proto"
+)
+
+// probeRunMetrics stores the cumulative metrics for probe runs for a target.
+// These metrics are sent via a data channel at the end of the probe run.
+type probeRunMetrics struct {
+	StorageSystem string
```
leczb: It's confusing that the type itself is not exported but its fields are.
Either make the type exported (if it's needed in other packages) or make its fields non-exported as well.

hermes/probe/metric.go
```diff
@@ -0,0 +1,107 @@
...
+	Target        string
+	BucketName    string
+
+	// probeOpLatency is used to record latency values with distinct labels
+	// per exit status per probe operation per target.
+	// Recommended usage: probeOpLatency[probeOp][exitStatus].Metric("latency").AddFloat64(<val>)
+	ProbeOpLatency map[string]map[string]*metrics.EventMetrics
```
leczb: Do these maps need mutexes?

hermes/probe/monitor.go
```diff
@@ -12,9 +12,164 @@
...
+
+	probepb "github.com/googleinterns/step224-2020/config/proto"
+	hermespb "github.com/googleinterns/step224-2020/hermes/proto"
+)
 
 // MonitorProbe holds aggregate information about all probe runs, per-target.
 // It also holds the config and options used to initialise the probe.
-type MonitorProbe struct{}
+type MonitorProbe struct {
+	name    string
```
leczb: It's surprising that `Config` is exported by `name` isn't.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,164 @@
...
+	Config  *probepb.HermesProbeDef
+	targets []*probepb.Target
+	opts    *options.Options
+
+	Hermes *hermes.Hermes
+
+	MetricMu sync.RWMutex
```
leczb: Is this mutex supposed to be expored? If so, it has to be documented (in general: all exported symbols must be documented).
Also, by convention this is called `mu` (if there's only one mutex).

hermes/probe/monitor.go
```diff
@@ -12,9 +12,164 @@
...
+//	- err: returns an error if there is an issue creating a distribution metric from the proto config passed.
+func (mp *MonitorProbe) initProbeMetrics() error {
+	mp.MetricMu.Lock()
+	defer mp.MetricMu.Unlock()
+	for _, t := range mp.targets {
+		if mp.LatencyMetrics[t.GetName()] != nil {
+			continue
+		}
+		var err error
+		if mp.LatencyMetrics[t.GetName()], err = NewProbeMetrics(*mp.Config, *t); err != nil {
```
leczb: Why dereference `mp.Config` and `t`?
I would change the signature of NewProbeMetrics to accept pointers instead.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,164 @@
...
+		}
+	}
+	return nil
+}
+
+// reportMetrics sends the metrics recorded in the last probe run to Cloudprober.
+// Arguments:
+//	- probeRun: metrics from a probe run on a target.
+//	- metricChan: metric channel passed from Cloudprober.
+func (mp *MonitorProbe) reportMetrics(prm probeRunMetrics, metricChan chan<- *metrics.EventMetrics) {
```
leczb: This can be a stand-alone function, as it doesn't use receiver (`mp`).

hermes/probe/monitor.go
```diff
@@ -12,9 +12,164 @@
...
+}
+
+// runProbe runs probe for all targets and update EventMetrics.
+// Arguments:
+//	- ctx: pass context to allow for complete cancellation of the probe.
+func (mp *MonitorProbe) runProbe(ctx context.Context, metricChan chan<- *metrics.EventMetrics) {
+	var wg sync.WaitGroup
+	for _, target := range mp.targets {
+		wg.Add(1)
+		go func(target probepb.Target) {
```
leczb: Again: change the signature to accept *probepb.Target.
In general: proto messages must be passed using pointers.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,164 @@
...
+		}(*target)
+	}
+	wg.Wait()
+}
+
+// runProbeForTarget runs the Hermes probing algorithm on a single target.
+// Arguments:
+//	- ctx: pass context to allow for cancellation of the probe.
+//	- target: the target to be probed
+func (mp *MonitorProbe) runProbeForTarget(ctx context.Context, target probepb.Target) (string, error) {
```
leczb: Also document what the return value is.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,164 @@
...
+	return "", nil
```
leczb: I would add a TODO or return a "unimplemented" error here.

hermes/probe/monitor_test.go
```diff
@@ -0,0 +1,123 @@
...
+	if err := mp.Init(wantName, wantOpts); err != nil {
+		t.Errorf("error when calling Init() on MonitorProbe object: wanted %v, got %v for error return value", nil, err)
+	}
+	if mp.name != wantName {
+		t.Errorf("error when calling Init() on MonitorProbe, wanted %v, got %v for name field", wantName, mp.name)
+	}
+	if !reflect.DeepEqual(mp.Config, wantConfig) {
+		t.Errorf("error when calling Init() on MonitorProbe, wanted %v, got %v for config field", wantConfig, mp.Config)
+	}
+	if !reflect.DeepEqual(mp.targets, wantConfig.GetTargets()) {
```
leczb: Not that this is wrong to have this check here, but it's preferable to test the *behaviour* of the package via its public API.
Its entirely possible that the `targets` field is initialized to the expected value, but then it's not *used* in the expected manner.

The problem with inspecting non-exported fields is that these are usually implementation details.
Ideally tests don't have to be changed when an internal implementation detail changes.
Imagine changing how the list of targets is stored: for example switching from string slice to a map.
Now you have to change the test as well, which is just busy-work.

leczb (changes requested):

hermes/probe/metric.go
```diff
@@ -0,0 +1,107 @@
...
+// Arguments:
+//	- conf: pass a HermesProbeDef config
+//	- target: pass the target for which metrics are to be collected.
+// Returns:
+//	- pm: returns an initialised *probeRunMetrics object.
+//	- err: returns an error if a latency distribution cannot be created from the config proto.
+func NewProbeMetrics(conf probepb.HermesProbeDef, target probepb.Target) (*probeRunMetrics, error) {
+	probeOps := []string{"total_probe_run", "check_nil", "read_and_verify", "delete_file", "create_file"}
+	apiCalls := []string{"list_files", "create_file", "delete_file", "get_file"}
+	probeExitStatuses := []string{"success", "op_timeout", "probe_failed", "file_missing", "bucket_missing", "file_corrupted", "file_read_failure", "file_metadata_mismatch", "unknown_file_found", "all_files_missing"}
```
cianbgoog: Ideally  put all these metric things into a var, in a seperate file, or at least var block at the top - this is likely to be the bit that people look for the most in the file, so it should be easy to find.

hermes/probe/metric.go
```diff
@@ -0,0 +1,107 @@
...
+
+// initLatencyDistMetric initialises a latency distribution metric with the given arguments.
+// Arguments:
+//	- latDistPb: pass the proto definition of a distribution metric.
+//	- ops: pass the list of operations associated with this metric.
+//	- exitStatuses: pass the list of possible exit statuses associated with this metric.
+// Returns:
+//	- latencyMetric: returns an initialises latency metric map.
+//	- err: returns an error if there a problem creating a distribution metric from the proto passed.
+func (p *probeRunMetrics) initLatencyDistMetric(latDistPb *metricpb.Dist, ops, exitStatuses []string) (map[string]map[string]*metrics.EventMetrics, error) {
```
cianbgoog: This is presumably only ever going to be called from one place? If so, I'd just inline it, as it's not massively complicated, and neither is the call site.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,164 @@
...
+)
 
 // MonitorProbe holds aggregate information about all probe runs, per-target.
 // It also holds the config and options used to initialise the probe.
-type MonitorProbe struct{}
+type MonitorProbe struct {
+	name    string
+	Config  *probepb.HermesProbeDef
+	targets []*probepb.Target
+	opts    *options.Options
```
cianbgoog: What is the difference between config & opts, and could you combine them?

cianbgoog (changes requested):

hermes/probe/metric.go
evanSpendlove: I can refactor these into enums and use a String() method when adding the label.

evanSpendlove: I will do that.

hermes/probe/metric.go
evanSpendlove: There is only a single instance of `probeRunMetrics` which has a mutex lock on it, so I don't think these fields need mutexes. If that's incorrect, please let me know and I will amend this.

hermes/probe/metric.go
```diff
@@ -0,0 +1,107 @@
...
+//	- conf: pass a HermesProbeDef config
+//	- target: pass the target for which metrics are to be collected.
+// Returns:
+//	- pm: returns an initialised *probeRunMetrics object.
+//	- err: returns an error if a latency distribution cannot be created from the config proto.
+func NewProbeMetrics(conf probepb.HermesProbeDef, target probepb.Target) (*probeRunMetrics, error) {
+	probeOps := []string{"total_probe_run", "check_nil", "read_and_verify", "delete_file", "create_file"}
+	apiCalls := []string{"list_files", "create_file", "delete_file", "get_file"}
+	probeExitStatuses := []string{"success", "op_timeout", "probe_failed", "file_missing", "bucket_missing", "file_corrupted", "file_read_failure", "file_metadata_mismatch", "unknown_file_found", "all_files_missing"}
+	apiExitStatuses := []string{"success", "api_call_failed", "file_missing", "bucket_missing", "file_corrupted", "file_read_failure", "file_metadata_mismatch", "unknown_file_found", "all_files_missing"}
```
deniadelcid: I think these would be better placed outside this function and probably as Enums rather than string arrays.

hermes/probe/metric.go
deniadelcid: uber nit: to make naming consistent with that of NewProbeMetrics, consider renaming `p` to `pm`.
```suggestion
func (pm *probeRunMetrics) initLatencyDistMetric(latDistPb *metricpb.Dist, ops, exitStatuses []string) (map[string]map[string]*metrics.EventMetrics, error) {
```

hermes/probe/monitor.go
deniadelcid:
```suggestion
		mp.LatencyMetrics[t.GetName()], err := NewProbeMetrics(*mp.Config, *t)
		if err != nil {
```

hermes/probe/monitor.go
```diff
@@ -12,9 +12,164 @@
...
+	hermespb "github.com/googleinterns/step224-2020/hermes/proto"
+)
 
 // MonitorProbe holds aggregate information about all probe runs, per-target.
 // It also holds the config and options used to initialise the probe.
-type MonitorProbe struct{}
+type MonitorProbe struct {
+	name    string
+	Config  *probepb.HermesProbeDef
+	targets []*probepb.Target
```
deniadelcid: Do you really need this? The `HermesProbeDef` already has a `targets` field that should make this already available to you: e.g.: you can get the list by referencing `Config.Targets`

hermes/probe/monitor.go
evanSpendlove: Ah okay, I understand. I will add a comment explaining why the cast is necessary.

hermes/probe/monitor.go
evanSpendlove: Config contains the proto configuration of the probe.
Opts contains various values and variables pulled/generated from the config.

Both are required for integration with Cloudprober. 

Even though we don't use opts internally, we need it as Cloudprober will use it.

deniadelcid: To piggy-back on this, I think you could combine them even further by making `Config` part of `hermes.Hermes`, since after all its a config that Hermes instances consume to make probe requests.

Having it here gives the impression that it's a config _for_ MonitorProbe, which isn't the case.

hermes/probe/monitor.go
evanSpendlove: Config is exported as it is used in other files in this package. There are several other fields such as `Hermes` and `Logger` that are used in other files and thus need to be exported. The `name` field isn't used much internally as it is primarily used by Cloudprober for storing probes in a map.

hermes/probe/monitor.go
evanSpendlove: It is supposed to be exported as these metrics are modified in other files within this package.

I've renamed it to `Mu` and added documentation as suggested.

hermes/probe/metric.go
evanSpendlove: It does not, I have renamed it to Metrics.

hermes/probe/metric.go
evanSpendlove: I have made this change as requested.

hermes/probe/metric.go
evanSpendlove: I have moved metrics to a `metrics` package.

hermes/probe/metric.go
evanSpendlove: That was a mistake on my part, thank you for catching it.
This should have been exported as well.

hermes/probe/monitor.go
evanSpendlove: I've amended several signatures to accept pointers instead, including this one.

hermes/probe/monitor.go
evanSpendlove: Okay, thank you. I'll amend this.

hermes/probe/monitor.go
evanSpendlove: Okay, I thought that pointers were more to be avoided. I'll keep this mind and amend this signature.

hermes/probe/monitor.go
evanSpendlove: Thanks, I've added the return value documentation.

hermes/probe/monitor.go
evanSpendlove: I've added a TODO for now.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,164 @@
...
+	"github.com/googleinterns/step224-2020/hermes"
+
+	probepb "github.com/googleinterns/step224-2020/config/proto"
+	hermespb "github.com/googleinterns/step224-2020/hermes/proto"
+)
 
 // MonitorProbe holds aggregate information about all probe runs, per-target.
 // It also holds the config and options used to initialise the probe.
-type MonitorProbe struct{}
+type MonitorProbe struct {
```
deniadelcid: It seems like in here you are collecting a lot of information that is already available in other places (e.g. HermesProbeDef already contains targets information). At the moment it is a bit confusing whether it operates at the target level (like your comment suggests) or the hermes.Hermes level. I think it would make sense to

1. Move/collapse/combine `Config`, `opts`, and `targets` to/into the Hermes struct. 
2. Since the intention is to run on a per-target basis, you could instead make a `Target` struct elsewhere that takes a `*probe.Target`, `LatencyMetric`, and mutex (plus whatever you think its relevant to monitor). Hermes would have a `[]*Targets` whose `*probe.Target` would either match the ones from Config, or be the exact same  thing.
3. At this stage you can decide whether you want to move the monitoring responsibilities to hermes.go or to keep it separate and have it take a Hermes struct as a field like you do now (in this file).

hermes/probe/monitor.go
```diff
@@ -12,9 +12,164 @@
...
+// Init initializes the probe with the given parameters.
+// This is a required method to implement the cloudprober.Probes.Probe interface.
+func (mp *MonitorProbe) Init(name string, opts *options.Options) error {
+	conf, ok := opts.ProbeConf.(*probepb.HermesProbeDef)
+	if !ok {
+		return fmt.Errorf("invalid argument: opts.ProbeConf is not of type *probepb.HermesProbeDef")
+	}
+	mp.name = name
+	mp.Config = conf
+	mp.targets = mp.Config.GetTargets()
```
deniadelcid: You could optimize the storage for mp by not duplicating information it already has. Remove this line, and simply iterate/pass around mp.Config.GetTargets(). 

Alternatively, if you aren't using any other member field from mp.Config, you can just have MetricProbe take just take in an array of targets.

deniadelcid (commented):
Some of the comments here might be a little contradicting. I just left them as I was reading along the file. Essentially, the biggest message here is to clarify the scope of `MetricProbe`. I would recommend to read through all of them and then decide the approach that makes most sense to you.

hermes/probe/metric.go
evanSpendlove: I've refactored these into consts using `iota` and a series of maps.

hermes/probe/metric.go
evanSpendlove: I've moved these into a series of consts and maps.

hermes/probe/metric.go
evanSpendlove: I've moved this functionality into the `NewMetrics()` function.

hermes/probe/metric.go
evanSpendlove: I've amended this by using consts and maps.

hermes/probe/metric.go
evanSpendlove: I've renamed this to `Metrics` and used `m`, which satisfies this too :)

hermes/probe/metric.go
evanSpendlove: I've renamed this accordingly to `NewMetric()`.
I've also gone through and removed mentions of unexported symbols in comments.

hermes/probe/metric.go
evanSpendlove: This is now out of date. These shouldn't need mutexes as there shouldn't be any concurrent access to these.

hermes/probe/monitor_test.go
evanSpendlove: Given that the Init() function only assigns values to other values and creates empty structs, would it be worthwhile removed the checks and just calling `Init()` and checking the error?

hermes/probe/monitor.go
evanSpendlove: This has been removed due to refactoring.

hermes/probe/monitor.go
evanSpendlove: I've removed this duplication due to refactoring.

hermes/probe/monitor.go
evanSpendlove: I've amended this to use a new `Target` struct which includes all of the information and state required for a probe run against a target. This also allows us to remove mutexes, which speeds up our code, by removing fields that required concurrent access.

hermes/probe/monitor.go
evanSpendlove: Thank you for raising this! I've thought about this and I think using a `Target` struct to encapsulate the information and state required for a target run is the best solution. It removes the need for a shared map of metrics per target and the duplication of the `mp.targets` field. It also allows us to store the state (via `StateJournal`) of the target bucket conveniently.

I think this also removes the need for a mutex lock as we no longer have fields that require concurrent access.

hermes/probe/metrics/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+import (
+	"fmt"
+	"time"
+
+	"github.com/google/cloudprober/metrics"
+
+	probepb "github.com/googleinterns/step224-2020/config/proto"
+)
+
+// ProbeOperation represents a possible probe operation metric label.
```
evanSpendlove, line 32:
```suggestion
// ProbeOperation is a probe operation metric label.
```

hermes/probe/metrics/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+type ProbeOperation int
+
+const (
+	// TOTAL_PROBE_RUN presents the metric label for the overall probe run.
```
evanSpendlove:
```suggestion
	// TOTAL_PROBE_RUN is the metric label for the overall probe run.
```

hermes/probe/metrics/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+		ALL_FILES_MISSING:      "all_files_missing",
+	}
+)
+
+// Metrics stores the cumulative metrics for probe runs for a target.
+// These metrics are sent via a data channel at the end of the probe run.
+type Metrics struct {
+	StorageSystem string
+	Target        string
+	BucketName    string
```
evanSpendlove: Data duplication - remove these as they are not needed.

hermes/probe/metrics/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+		FILE_CORRUPTED:         "file_corrupted",
+		FILE_READ_FAILURE:      "file_read_failure",
+		FILE_METADATA_MISMATCH: "file_metadata_mismatch",
+		UNKNOWN_FILE_FOUND:     "unknown_file_found",
+		ALL_FILES_MISSING:      "all_files_missing",
+	}
+)
+
+// Metrics stores the cumulative metrics for probe runs for a target.
+// These metrics are sent via a data channel at the end of the probe run.
```
evanSpendlove: Delete this line.
```suggestion
```

hermes/probe/metrics/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+type Metrics struct {
+	StorageSystem string
+	Target        string
+	BucketName    string
+
+	// probeOpLatency is used to record latency values with distinct labels
+	// per exit status per probe operation per target.
+	// Recommended usage: probeOpLatency[probeOp][exitStatus].Metric("latency").AddFloat64(<val>)
```
evanSpendlove: Add enum/const to example.
```suggestion
	// Recommended usage: probeOpLatency[ProbeOperation][ExitStatus].Metric("latency").AddFloat64(<val>)
```

hermes/probe/metrics/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+	ProbeOpLatency map[string]map[string]*metrics.EventMetrics
+	// apiCallLatency is used to record latency values with distinct labels
+	// per exit status per API call per target.
+	// Recommended usage: apiCallLatency[apiCall][exitStatus].Metric("latency").AddFloat64(<val>)
```
evanSpendlove:
```suggestion
	// Recommended usage: apiCallLatency[ApiCall][ExitStatus].Metric("latency").AddFloat64(<val>)
```

hermes/probe/metrics/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+	ApiCallLatency map[string]map[string]*metrics.EventMetrics
+}
+
+// NewMetrics creates a new *probeRunMetrics object and initialises the fields inside it.
```
evanSpendlove:
```suggestion
// NewMetrics creates a new *Metrics object and initialises the fields inside it.
```

hermes/probe/metrics/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+//	- conf: pass a HermesProbeDef config
+//	- target: pass the target for which metrics are to be collected.
+// Returns:
+//	- m: returns an initialised *probeRunMetrics object.
+//	- err: returns an error if a latency distribution cannot be created from the config proto.
+func NewMetrics(conf *probepb.HermesProbeDef, target *probepb.Target) (*Metrics, error) {
+	m := &Metrics{
+		StorageSystem: target.GetTargetSystem().String(),
+		Target:        target.GetName(),
+		BucketName:    target.GetBucketName(),
```
evanSpendlove: Remove these.

hermes/probe/metrics/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+	m.ProbeOpLatency = make(map[string]map[string]*metrics.EventMetrics, len(ProbeOperations))
```
evanSpendlove: Move this into `&m{}`.

hermes/probe/metrics/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+			m.ProbeOpLatency[op][e] = metrics.NewEventMetrics(time.Now()).
+				AddMetric("hermes_probe_operation_latency_s", probeOpLatDist.Clone()).
+				AddLabel("storage_system", m.StorageSystem).
+				AddLabel("target", fmt.Sprintf("%s:%s", m.Target, m.BucketName)).
+				AddLabel("probe_operation_type", op).
+				AddLabel("exit_status", e)
+		}
+	}
+
+	m.ApiCallLatency = make(map[string]map[string]*metrics.EventMetrics, len(ApiCalls))
```
evanSpendlove: Move this into `&m{}`.

hermes/probe/metrics/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("invalid argument: error creating probe latency distribution from the specification (%v): %w", conf.GetProbeLatencyDistribution(), err)
+	}
+
+	for _, op := range ProbeOperations {
+		m.ProbeOpLatency[op] = make(map[string]*metrics.EventMetrics, len(ExitStatuses))
+		for _, e := range ExitStatuses {
+			m.ProbeOpLatency[op][e] = metrics.NewEventMetrics(time.Now()).
+				AddMetric("hermes_probe_operation_latency_s", probeOpLatDist.Clone()).
+				AddLabel("storage_system", m.StorageSystem).
```
evanSpendlove: Amend to use args.

hermes/probe/metrics/metrics.go
```diff
@@ -0,0 +1,192 @@
...
+		return nil, fmt.Errorf("invalid argument: error creating probe latency distribution from the specification (%v): %v", conf.GetApiCallLatencyDistribution(), err)
+	}
+
+	for _, op := range ApiCalls {
+		m.ApiCallLatency[op] = make(map[string]*metrics.EventMetrics, len(ExitStatuses))
+		for _, e := range ExitStatuses {
+			m.ApiCallLatency[op][e] = metrics.NewEventMetrics(time.Now()).
+				AddMetric("hermes_probe_operation_latency_s", apiCallLatDist.Clone()).
+				AddLabel("storage_system", m.StorageSystem).
+				AddLabel("target", fmt.Sprintf("%s:%s", m.Target, m.BucketName)).
```
evanSpendlove: Amend to use args.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,161 @@
...
+
+	probepb "github.com/googleinterns/step224-2020/config/proto"
+	journalpb "github.com/googleinterns/step224-2020/hermes/proto"
+)
+
+// Target holds all of the required information and state for a given target run.
+type Target struct {
+	target *probepb.Target
+
+	// Journal stores the state of MonitorProbe as a combination of a next operation intent enum and a filenames map
```
evanSpendlove:
```suggestion
	// Stores the state of MonitorProbe as a combination of a next operation intent enum and a filenames map.
```

hermes/probe/monitor.go
```diff
@@ -12,9 +12,161 @@
...
+	journal *journalpb.StateJournal
+
+	// latencyMetrics stores the api call and probe operation latency for a given target run.
```
evanSpendlove:
```suggestion
	// Stores the api call and probe operation latency for a given target run.
```

hermes/probe/monitor.go
```diff
@@ -12,9 +12,161 @@
...
+	latencyMetrics *metrics.Metrics
+}
 
 // MonitorProbe holds aggregate information about all probe runs, per-target.
 // It also holds the config and options used to initialise the probe.
-type MonitorProbe struct{}
+type MonitorProbe struct {
+	name string
+	// Config holds the configuration details for MonitorProbe.
+	Config  *probepb.HermesProbeDef
```
evanSpendlove: Should be able to keep fields private by passing them in function calls.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,161 @@
...
+		mp.targets = append(mp.targets, run)
+	}
+	mp.opts = opts
+	mp.Logger = opts.Logger
+
+	return nil
+}
+
+// Start runs the probe indefinitely, unless cancelled, at the configured interval.
+// Probe metrics will be sent via the metricChan at the beginning of the next probe run.
```
evanSpendlove: Update comment.
```suggestion
// Probe metrics will be sent via the metricChan at the end of a probe run.
```

hermes/probe/monitor.go
```diff
@@ -12,9 +12,161 @@
...
+		case <-ctx.Done():
+			probeTicker.Stop()
+			return
+		case <-probeTicker.C:
+			mp.runProbe(ctx, metricChan)
+		}
+	}
+}
+
+// reportMetrics sends the metrics recorded in the last probe run to Cloudprober.
```
evanSpendlove:
```suggestion
// reportMetrics sends the metrics recorded in the current probe run to Cloudprober.
```

hermes/probe/monitor.go
evanSpendlove: Actually, these need to be exported, so I've update them accordingly.

hermes/probe/monitor.go
evanSpendlove: Actually, these need to be exported, so I've update them accordingly.

hermes/probe/monitor.go
evanSpendlove: Ack.

hermes/probe/monitor.go
evanSpendlove: Ack.

hermes/probe/monitor.go
evanSpendlove: Ack.

hermes/probe/metrics/metrics.go
evanSpendlove: Ack.

hermes/probe/metrics/metrics.go
evanSpendlove: Ack.

hermes/probe/metrics/metrics.go
evanSpendlove: Removed.

hermes/probe/metrics/metrics.go
evanSpendlove: Removed.

hermes/probe/metrics/metrics.go
evanSpendlove: Done.

hermes/probe/metrics/metrics.go
evanSpendlove: Done.

hermes/probe/metrics/metrics.go
evanSpendlove: Done.

hermes/probe/metrics/metrics.go
evanSpendlove: Done.

hermes/probe/metrics/metrics.go
evanSpendlove: Ack.

hermes/probe/metrics/metrics.go
evanSpendlove: Ack.

hermes/probe/metrics/metrics.go
evanSpendlove: Ack.

cianbgoog (approved):

hermes/probe/monitor.go
```diff
@@ -12,9 +12,160 @@
...
+	journalpb "github.com/googleinterns/step224-2020/hermes/proto"
+)
+
+// Target holds all of the required information and state for a given target run.
+type Target struct {
+	// Target stores the proto config for the target to be probed.
+	Target *probepb.Target
+
+	// Journal stores the state of MonitorProbe as a combination of a next operation intent enum and a filenames map.
+	Journal *journalpb.StateJournal
```
deniadelcid: It seems that Hermes also takes in a `*journalpb.StateJournal`. How is this one different from the one in Hermes? Also, if each target has its own state journal now, does it still make sense for Hermes to have one?

hermes/probe/monitor_test.go
leczb: I would do exactly that: call Init() and check there's no error.
In other tests, Init() will be part of the set-up. If those tests pass, there's implicit confirmation that Init() did its job.

hermes/probe/metrics/metrics.go
```diff
@@ -0,0 +1,183 @@
...
+	FILE_METADATA_MISMATCH
+	// UNKNOWN_FILE_FOUND indicates that an unknown file, not created by Hermes, was found in the target bucket.
+	UNKNOWN_FILE_FOUND
+	// ALL_FILES_MISSING indicates that all of the Hermes files were missing.
+	ALL_FILES_MISSING
+)
+
+var (
+	// ProbeOps maps ProbeOperation constants to their metric label string equivalent.
+	ProbeOps = map[ProbeOperation]string{
```
leczb: Optional: I would call this map ProbeOpName.
Same for the other maps.

hermes/probe/metrics/metrics.go
```diff
@@ -0,0 +1,183 @@
...
+		ALL_FILES_MISSING:      "all_files_missing",
+	}
+)
+
+// Metrics stores the cumulative metrics for probe runs for a target.
+type Metrics struct {
+	// probeOpLatency is used to record latency values with distinct labels
+	// per exit status per probe operation per target.
+	// Recommended usage: probeOpLatency[ProbeOperation][ExitStatus].Metric("latency").AddFloat64(<val>)
+	ProbeOpLatency map[string]map[string]*metrics.EventMetrics
```
leczb: Consider using the numerical identifiers as map keys and translating them to strings at the very last stage.

hermes/probe/metrics/metrics.go
```diff
@@ -0,0 +1,183 @@
...
+		ProbeOpLatency: make(map[string]map[string]*metrics.EventMetrics, len(ProbeOps)),
+		ApiCallLatency: make(map[string]map[string]*metrics.EventMetrics, len(ApiCalls)),
+	}
+
+	probeOpLatDist, err := metrics.NewDistributionFromProto(conf.GetProbeLatencyDistribution())
+	if err != nil {
+		return nil, fmt.Errorf("invalid argument: error creating probe latency distribution from the specification (%v): %w", conf.GetProbeLatencyDistribution(), err)
+	}
+
+	for _, op := range ProbeOps {
```
leczb: If you switch to using the numeric identifier in the map, this will change to
```
for op := range ProbeOps {
```

hermes/probe/monitor.go
```diff
@@ -12,9 +12,160 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+// Authors:
+// - Alicja Kwiecinska (kwiecinskaa@google.com) github: alicjakwie
+// - Evan Spendlove, GitHub: evanSpendlove.
+//
 // Monitor implements a probe that monitors a storage system using the
 // Hermes algorithm.
+// Package probe implements the probe that Hermes uses to monitor
```
leczb: GoDoc expects the "// Package probe ..." on the first line in the file.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,160 @@
...
+// a storage system.
+
```
leczb: Delete this empty line to attach the package comment to "package probe".

hermes/probe/monitor.go
leczb: I agree that as a reader I'm puzzled by the presence of both "config" and "opts".

hermes/probe/monitor.go
```diff
@@ -12,9 +12,160 @@
...
+	// Cast from Cloudprober defined interface to HermesProbeDef.
+	// This allows for accessing the variables and methods of a HermesProbeDef object.
+	conf, ok := opts.ProbeConf.(*probepb.HermesProbeDef)
+	if !ok {
+		return fmt.Errorf("invalid argument: opts.ProbeConf is not of type *probepb.HermesProbeDef")
+	}
+	mp.name = name
+	mp.config = conf
+
+	mp.targets = make([]*Target, 0)
```
leczb: There's no need to initialize a slice: append() works correctly with a nil value and does the initial allocation implicitly.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,160 @@
...
+	for _, t := range mp.config.GetTargets() {
+		run := &Target{
```
leczb: No need for the intermediate variable.
Just
```
m.targets = append(m.targets, &Target{
...
})

File: hermes/probe/monitor.go [hunk 78]
```diff
@@ -12,9 +12,160 @@
...
+	}
+
+	for _, call := range run.ApiCallLatency {
+		for _, m := range call {
+			metricChan <- m
+		}
+	}
+}
+
+// runProbe runs probe for all targets and update EventMetrics.
```
leczb: Update this comment.

hermes/probe/monitor.go
```diff
@@ -12,9 +12,160 @@
...
+	for _, t := range mp.targets {
+		wg.Add(1)
+		go func(t *Target) {
+			defer wg.Done()
+			var err error
+			if t.LatencyMetrics, err = metrics.NewMetrics(mp.config, t.Target); err != nil {
+				mp.logger.Errorf(err.Error())
+				return
+			}
+			probeCtx, _ := context.WithDeadline(ctx, time.Now().Add(mp.opts.Timeout)) //TODO: Change to mp.Config. ...
```
leczb: Nit: Avoid in-line comments in general. Also: add space after "//".

hermes/probe/monitor.go
```diff
@@ -12,9 +12,160 @@
...
+			start := time.Now()
+			exitStatus, err := mp.runProbeForTarget(probeCtx, t)
+			if err != nil {
+				mp.logger.Errorf(err.Error())
+				t.LatencyMetrics.ProbeOpLatency[metrics.ProbeOps[metrics.TOTAL_PROBE_RUN]][metrics.ExitStatuses[exitStatus]].Metric("latency").AddFloat64(time.Now().Sub(start).Seconds())
```
leczb: Again, using the numeric ID in the maps would remove the need for this translation.

leczb (changes requested):

hermes/probe/monitor_test.go
evanSpendlove: Excellent! Thank you, I'll go with this approach.

hermes/probe/monitor.go
evanSpendlove: Ah okay, I think I misunderstood this before. I'll amend this.

hermes/probe/monitor.go
evanSpendlove: Ack.

hermes/probe/monitor.go
evanSpendlove: Oh that's neat, I'll correct this.

hermes/probe/monitor.go
evanSpendlove: I've amended this.

hermes/probe/monitor.go
evanSpendlove: Updated to remove unexported symbol mention and to be more descriptive.

hermes/probe/monitor.go
evanSpendlove: Okay, thank you. I've completed this TODO and removed the comment.

hermes/probe/metrics/metrics.go
evanSpendlove: I've renamed these to "_Name"

hermes/probe/metrics/metrics.go
evanSpendlove: This is a great change, thank you! I've refactored accordingly.

hermes/probe/metrics/metrics.go
evanSpendlove: Ack.

hermes/probe/monitor.go
evanSpendlove: Ack.

hermes/probe/monitor.go
leczb: I don't see the fix.

-- REPLY --
It shows up for me... I think it's GitHub being strange..

hermes/probe/monitor.go
```diff
@@ -110,7 +108,7 @@ func (mp *MonitorProbe) Start(ctx context.Context, metricChan chan<- *cpmetrics.
 	}
 }
 
-// reportMetrics sends the metrics recorded in the current probe run to Cloudprober.
+// Sends the metrics recorded in the current probe run to Cloudprober.
```
leczb: GoDoc expects the first word of the comment to match the identifier it belongs to.
Here and elsewhere.

hermes/probe/monitor.go
```diff
@@ -143,22 +142,22 @@ func (mp *MonitorProbe) runProbe(ctx context.Context, metricChan chan<- *cpmetri
 				mp.logger.Errorf(err.Error())
 				return
 			}
-			probeCtx, _ := context.WithDeadline(ctx, time.Now().Add(mp.opts.Timeout)) //TODO: Change to mp.Config. ...
+			probeCtx, _ := context.WithDeadline(ctx, time.Now().Add(time.Duration(mp.config.GetIntervalSec())))
```
leczb: The time.Duration unit is nanoseconds. To fix this, multiple `mp.config.GetIntervalSec()` by `time.Second`.
I expect this bug would have been caught by unit-tests, as the operation would always time out.

leczb (changes requested):

hermes/probe/monitor.go
evanSpendlove: Ack.

hermes/probe/monitor.go
evanSpendlove: Ah thank you for pointing this out. Currently, the unit tests are quite sparse. I have a TODO to add more once the code is satisfactory.

hermes/probe/monitor.go
```diff
@@ -95,7 +95,7 @@ func (mp *MonitorProbe) Init(name string, opts *options.Options) error {
 //	- ctx: context provided for cancelling probe.
 //	- metricChan: unidirectional channel used for sending metrics to be surfaced.
 func (mp *MonitorProbe) Start(ctx context.Context, metricChan chan<- *cpmetrics.EventMetrics) {
-	probeTicker := time.NewTicker(mp.opts.Interval)
+	probeTicker := time.NewTicker(time.Duration(mp.config.GetIntervalSec()) * time.Second)
```
leczb: I would add a method to MonitorProbe that returned the execution interval as a time.Duration value.

leczb (approved):
LGTM

hermes/probe/monitor.go
```diff
@@ -12,9 +12,160 @@
...
+
+	// LatencyMetrics stores the api call and probe operation latency for a given target run.
+	// Metrics are stored with additional labels to record operation type and exit status.
+	LatencyMetrics *metrics.Metrics
+}
 
 // MonitorProbe holds aggregate information about all probe runs, per-target.
 // It also holds the config and options used to initialise the probe.
-type MonitorProbe struct{}
+type MonitorProbe struct {
```
deniadelcid: nit: MonitorProbe sounds like an action, maybe rename it to ProbeMonitor or just Prober.

deniadelcid (approved):

hermes/probe/monitor.go
evanSpendlove: Ack.

hermes/probe/monitor.go
evanSpendlove: I have removed the Hermes.go file, as it only contains the `*journalpb.StateJournal` and this is better placed in the `Target` struct.

hermes/probe/monitor.go
evanSpendlove: This has been renamed to `Probe`, and monitor.go is now `probe.go`.

hermes/probe/probe.go
```diff
@@ -51,19 +51,33 @@ type Target struct {
...
-type MonitorProbe struct {
+type Probe struct {
 	name    string
 	config  *probepb.HermesProbeDef
 	targets []*Target
 	opts    *options.Options
 	logger  *logger.Logger
 }
 
+// getIntervalSec returns the probing interval in seconds as a time.Duration.
```
leczb: Avoid the "get" prefix in general. How about "probeInterval" or just "interval".
Also, the "Sec" in the name is misleading, as it returns time.Duration not seconds.

hermes/probe/probe.go
```diff
@@ -51,19 +51,33 @@ type Target struct {
...
+// Returns:
+//	- time.Duration: returns the probing interval in seconds
+func (mp *Probe) getIntervalSec() time.Duration {
+	return time.Duration(mp.config.GetIntervalSec()) * time.Second
+}
+
+// getTimeoutSec returns the probe timeout in seconds as a time.Duration.
+// Returns:
+//	- time.Duration: returns the probe timeout in seconds
+func (mp *Probe) getTimeoutSec() time.Duration {
```
leczb: Same as above.

leczb (approved):
LGTM once these small changes are made.

alicjakwie (approved):

hermes/probe/probe.go
evanSpendlove: Okay, thanks, I've amended these.

hermes/probe/probe.go
evanSpendlove: Ack.

PR #61, opened 2020-09-14
Create delete_file.go and test file.

evanSpendlove:


hermes/deletionPicker.go
```diff
@@ -28,9 +27,8 @@ const (
...
 )
 
-// PickFileToDelete picks which file to delete and returns a string: file name in the form "Hermes_ID".
-func PickFileToDelete() string {
+// PickFileToDelete picks which file to delete and returns the integer ID of this file.
+func PickFileToDelete() int {
 	rand.Seed(time.Now().UnixNano())
-	fileID := rand.Intn(numberOfDeletableFiles) + beg // rand.Intn will return a natural number in the range [0, number_of_deletable_files) so fileID will be in the range [beg, number_of_deletable_files)
-	return fmt.Sprintf("Hermes_%02d", fileID)
+	return rand.Intn(numberOfDeletableFiles) + beg // rand.Intn will return a natural number in the range [0, number_of_deletable_files) so fileID will be in the range [beg, number_of_deletable_files)
```
cianbgoog: We don't need this comment.

hermes/probe/delete_file.go
```diff
@@ -0,0 +1,104 @@
...
+//		- [...].bucket_missing: the target bucket on this target system was not found.
+//		- [...].probe_failed: there was an error during one of the API calls and the probe failed.
+//		- [...].deleted_file_found: the file was deleted but it is still found in the target bucket.
+//		- [...].list_bucket_failed: the listBucket operation failed when checking if the target file was deleted.
+func DeleteRandomFile(ctx context.Context, config *probepb.HermesProbeDef, target *Target, client *stiface.Client, logger *logger.Logger) (int, error) {
+	bucket := target.Target.GetBucketName()
+
+	fileID := hermes.PickFileToDelete()
+
+	filename, ok := target.Journal.Filenames[int32(fileID)]
```
cianbgoog: Why not just make `PickFileToDelete()` return an `int32, or `Filenames` contain whatever `PickFilesToDelete()` returns?

hermes/probe/delete_file.go
```diff
@@ -0,0 +1,104 @@
...
+	if !ok {
+		return fileID, fmt.Errorf("deleteRandomFile(id: %d).hermes_file_missing: could not delete file %s, file not found", fileID, filename)
```
cianbgoog: Given that you'll need to know what kind of error this is, you may want a custom error object, here and elsewhere.

hermes/probe/delete_file.go
```diff
@@ -0,0 +1,104 @@
...
+			status = m.ExitStatuses[m.PROBE_FAILED]
+		}
+		target.LatencyMetrics.ApiCallLatency[m.ApiCalls[m.API_DELETE_FILE]][status].Metric("latency").AddFloat64(time.Now().Sub(start).Seconds())
+		return fileID, fmt.Errorf("deleteRandomFile(id: %d).%s: could not delete file %s: %w", fileID, status, filename, err)
+	}
+	target.LatencyMetrics.ApiCallLatency[m.ApiCalls[m.API_DELETE_FILE]][m.ExitStatuses[m.SUCCESS]].Metric("latency").AddFloat64(time.Now().Sub(start).Seconds())
+
+	query := &storage.Query{Prefix: filename}
+	start = time.Now()
+	it := (*client).Bucket(bucket).Objects(ctx, query)
```
cianbgoog: Please pick a more descriptive name than `it` - `it` is especially bad given how it's semantically meaningful in at least some languages.

hermes/probe/delete_file.go
```diff
@@ -0,0 +1,104 @@
...
+	for {
+		obj, err := it.Next()
+		if obj.Name == filename {
+			return fileID, fmt.Errorf("deleteRandomFile(id %d).deleted_file_found: object %v in bucket %q still listed after delete", fileID, *obj, bucket)
+		}
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			return fileID, fmt.Errorf("deleteRandomFile(id: %d).list_bucket_failed: unable to list bucket %q: %w", fileID, bucket, err)
```
cianbgoog: Maybe look at error wrapping here: https://peter.bourgon.org/blog/2019/09/11/programming-with-errors.html

hermes/probe/delete_file.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+	}
+
+	// Update in-memory NIL file after delete operation.
+	delete(target.Journal.Filenames, int32(fileID))
```
cianbgoog: You should add a TODO to update the actual NIL file, or update this comment.

hermes/probe/delete_file_test.go
```diff
@@ -0,0 +1,55 @@
...
+	if err := mp.Init("testMonitorProbe1", opts); err != nil {
+		t.Errorf("error when calling Init() on MonitorProbe object: wanted %v, got %v for error return value", nil, err)
+	}
+
+	fileID, err := DeleteRandomFile(ctx, mp.config, mp.targets[0], &client, mp.logger)
+	if err != nil {
+		t.Errorf("deleteRandomFile(ID: %d) failed: expected error as %v, got %v", fileID, nil, err)
+	}
+
+	// w := client.Bucket("my-bucket").Object("my-object").NewWriter(ctx)
```
cianbgoog: Delete this, rather than commenting it out - it'll still be in your git history anyway

cianbgoog (changes requested):

hermes/probe/delete_file.go
evanSpendlove: I was planning on updating the NIL file within the algorithm in RunProbeForTarget() rather than in the DeleteRandomFile() function. Is that okay?

hermes/probe/delete_file.go
```diff
@@ -0,0 +1,104 @@
...
+
+// deleteRandomFile deletes a random file in the target storage system bucket.
+// It then checks that the file has been deleted by trying to get the object.
+// Arguments:
+// TODO(evanSpendlove): Update comment here
+//	- ctx: pass the context so this probe can be cancelled if needed.
+//	- config: pass the HermesProbeDef config for the probe calling this function.
+//	- target: pass the target run information.
+//	- client: pass an initialised storage client for this target system.
+//	- logger: pass the logger associated with the probe calling this function.
```
deniadelcid: Might be helpful to note that this is a Google Cloud Logging logger.

hermes/probe/delete_file.go
```diff
@@ -0,0 +1,104 @@
...
+// Returns:
+//	- fileID: returns the ID of the file delete OR a missing file to be created if one is found.
+//	- err:
+//		- [...].hermes_file_missing: the file to be deleted does not exist in Hermes' StateJournal.
```
deniadelcid: I think you can remove the "hermes" portion and leave this just as "not_found". Also, would the file be truly missing in this case? Is it possible for the file to be present in the storage system and not in the state journal?

hermes/probe/delete_file.go
```diff
@@ -0,0 +1,104 @@
...
+		return fileID, fmt.Errorf("deleteRandomFile(id: %d).hermes_file_missing: could not delete file %s, file not found", fileID, filename)
+	}
+
+	file := (*client).Bucket(bucket).Object(filename)
+
+	start := time.Now()
+	if err := file.Delete(ctx); err != nil {
+		var status string
+		if err == storage.ErrObjectNotExist {
+			status = m.ExitStatuses[m.FILE_MISSING]
```
deniadelcid: nit: You could make this a switch instead.

hermes/probe/delete_file.go
```diff
@@ -0,0 +1,104 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Delete_file implements the probe operation for deleting a file in a
+// storage system.
+// Package probe implements the probe that Hermes uses to monitor
```
leczb: "// Package probe" must be the first line in the file.
Also note that this package comment only has to be present in one of the files belonging to the package.
The typical pattern is to have a single file per package with the file named exactly like the package.

hermes/probe/delete_file.go
```diff
@@ -0,0 +1,104 @@
...
+//		- [...].hermes_file_missing: the file to be deleted does not exist in Hermes' StateJournal.
+//		- [...].file_missing: the file to be deleted could not be found in the target bucket.
+//		- [...].bucket_missing: the target bucket on this target system was not found.
+//		- [...].probe_failed: there was an error during one of the API calls and the probe failed.
+//		- [...].deleted_file_found: the file was deleted but it is still found in the target bucket.
+//		- [...].list_bucket_failed: the listBucket operation failed when checking if the target file was deleted.
+func DeleteRandomFile(ctx context.Context, config *probepb.HermesProbeDef, target *Target, client *stiface.Client, logger *logger.Logger) (int, error) {
+	bucket := target.Target.GetBucketName()
+
+	fileID := hermes.PickFileToDelete()
```
leczb: It's surprising that the logic for picking the file is in another package. Why not put that code here?

hermes/probe/delete_file.go
leczb: "file not found" is a bit misleading. It suggests that the *storage system* reports the file as "not found".
I would change it to say "unknown" or invalid.
Also, it would be avoidable if PickFileToDelete() took the list of valid filenames as an input and returned one of them.

hermes/probe/delete_file.go
```diff
@@ -0,0 +1,104 @@
...
+
+	filename, ok := target.Journal.Filenames[int32(fileID)]
+	if !ok {
+		return fileID, fmt.Errorf("deleteRandomFile(id: %d).hermes_file_missing: could not delete file %s, file not found", fileID, filename)
+	}
+
+	file := (*client).Bucket(bucket).Object(filename)
```
leczb: It's surprising the `client` is dereferenced.
This should be just `client.Bucket(bucket).Object(filename)`.

hermes/probe/delete_file.go
```diff
@@ -0,0 +1,104 @@
...
+
+	start := time.Now()
+	if err := file.Delete(ctx); err != nil {
+		var status string
+		if err == storage.ErrObjectNotExist {
```
leczb: Not that this is wrong, but the `switch` statement is better suited for this.

hermes/probe/delete_file.go
```diff
@@ -0,0 +1,104 @@
...
+			status = m.ExitStatuses[m.FILE_MISSING]
+		} else if err == storage.ErrBucketNotExist {
+			status = m.ExitStatuses[m.BUCKET_MISSING]
+		} else {
+			status = m.ExitStatuses[m.PROBE_FAILED]
+		}
+		target.LatencyMetrics.ApiCallLatency[m.ApiCalls[m.API_DELETE_FILE]][status].Metric("latency").AddFloat64(time.Now().Sub(start).Seconds())
+		return fileID, fmt.Errorf("deleteRandomFile(id: %d).%s: could not delete file %s: %w", fileID, status, filename, err)
```
leczb: Nit: Suggested error message "Delete(%q, %q) failed; status: %s: %w", bucket, filename, status, err

hermes/probe/delete_file.go
```diff
@@ -0,0 +1,104 @@
...
+		if err != nil {
+			return fileID, fmt.Errorf("deleteRandomFile(id: %d).list_bucket_failed: unable to list bucket %q: %w", fileID, bucket, err)
+		}
+	}
+
+	// Update in-memory NIL file after delete operation.
+	delete(target.Journal.Filenames, int32(fileID))
+
+	logger.Infof("Object %v deleted in bucket %s.", file, bucket)
+	return fileID, nil
```
leczb: What's the purpose of returning the file ID?

hermes/probe/delete_file_test.go
```diff
@@ -0,0 +1,55 @@
...
+	mp := &MonitorProbe{}
+
+	_, cfg := GenTestConfig("testMonitorProbe1")
+	opts := GenOptsFromConfig(t, cfg)
+	if err := mp.Init("testMonitorProbe1", opts); err != nil {
+		t.Errorf("error when calling Init() on MonitorProbe object: wanted %v, got %v for error return value", nil, err)
+	}
+
+	fileID, err := DeleteRandomFile(ctx, mp.config, mp.targets[0], &client, mp.logger)
+	if err != nil {
```
leczb: Also verify that the file was deleted.

leczb (changes requested):

hermes/probe/delete_file.go
deniadelcid: If that's the case, is this comment still applicable?

-- REPLY --
I think so. This comment describes updating the in-memory NIL file represented as the Journal struct. The actual NIL file stored on GCS is updated elsewhere.

hermes/probe/delete_file_test.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+func TestDeleteRandomFile(t *testing.T) {
+	ctx := context.Background()
+	c, err := storage.NewClient(ctx)
+	if err != nil {
+		t.Fatalf("client conn failed: could not connect to storage system with client: %v", err)
+	}
+
+	client := stiface.AdaptClient(c) // Don't use this, use fakeClient.
```
deniadelcid: Should this comment be a TODO instead?

hermes/probe/delete_file_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"cloud.google.com/go/storage"
+	"github.com/googleapis/google-cloud-go-testing/storage/stiface"
+)
+
+func TestDeleteRandomFile(t *testing.T) {
```
deniadelcid: Although there isn't support for NIL files, yet. I think it might be worth adding a test scenario to delete a nil file.

hermes/probe/delete_file.go
evanSpendlove: I think so. This comment describes updating the in-memory NIL file represented as the Journal struct. The actual NIL file stored on GCS is updated elsewhere.

hermes/probe/delete_file.go
evanSpendlove: I understand why this is misleading. The reason that I have this check here is to identify if there is an inconsistency in the in-memory state of Hermes (i.e. Journal struct). This should never be triggered unless the Journal is not up-to-date, as there should always be 50 entries in the Filenames map.

hermes/probe/delete_file.go
evanSpendlove: The ID is returned as it is used in the next function in the algorithm, createFile(). Rather than having to search for the one missing file ID in createFile(), I thought it would be better to return it and pass it to createFile().

hermes/probe/delete_file_test.go
evanSpendlove: Yes, this will be added now that mocks have been added to Alicja's create_file_test.go

hermes/probe/delete_file.go
evanSpendlove: The reason that I titled this as "hermes_file_missing" is because the file is not found in Hermes' state, but it could exist on the storage system if the Journal is inconsistent due to some error/bug when updating the Journal.

hermes/probe/delete_file_test.go
evanSpendlove: Since there isn't any variable passed to identify the file to be deleted and the NIL file won't ever be selected by the number returned from DeletionPicker.go, I don't think we can add a test scenario of deleting a NIL file.

evanSpendlove: As in, the NIL file won't be chosen to be deleted, and we can't modify the `DeleteRandomFile()` function inside the test to do this, so I'm not sure how we could test this.

hermes/deletionPicker.go
evanSpendlove: Ack.

hermes/probe/delete_file.go
evanSpendlove: I've moved `PickFileToDelete()` into delete_file.go and modified the return value to `int32`.

hermes/probe/delete_file.go
evanSpendlove: Ah apologies for this. `it` was used in the godoc for the storage library, so I kept it. I've changed this to `objects` as it is an iterator over objects.

hermes/probe/delete_file_test.go
evanSpendlove: Ack.

hermes/probe/delete_file.go
evanSpendlove: I've moved this from deletionPicker.go into this file, and deleted the previous file.

hermes/probe/delete_file.go
evanSpendlove: I've moved this into a `delete` package and adjusted the package comment appropriately.

hermes/probe/delete_file.go
evanSpendlove: I've changed this to a switch statement.

hermes/probe/delete_file.go
evanSpendlove: Ack.

hermes/probe/delete_file_test.go
evanSpendlove: Yep, thank you.

hermes/probe/delete_file.go
evanSpendlove: Ack.

hermes/probe/delete_file.go
evanSpendlove: We're passing a pointer to the client interface defined in the Google Cloud mocking library, so we need to dereference it to call its methods. It is quite strange 😕 

Source: https://github.com/googleapis/google-cloud-go-testing/tree/f550565525113d92f31e16591986917006f86eb9

leczb: Interfaces are already pointers, so there's no point in storing a pointer to an interface type.
Change the type of the client arg from `*stiface.Client` to `stiface.Client` and remove the `*` from here (and elsewhere).

evanSpendlove: Thank you for clarifying this. I've amended the argument and updated the code.

hermes/probe/delete_file.go
evanSpendlove: I've amended the error message to match this style.

hermes/probe/delete_file.go
evanSpendlove: I've created a new PR to add a ProbeError object that we can use to return custom errors with statuses. When this is merged, I will add it to this file.

hermes/probe/delete_file.go
evanSpendlove: I've created a new PR to add a ProbeError object that we can use to return custom errors with statuses. When this is merged, I will add it to this file.

hermes/probe/delete/delete_file.go
```diff
@@ -0,0 +1,138 @@
...
+	logger.Infof("Object %v deleted in bucket %s.", file, bucket)
+	return fileID, nil
+}
+
+// pickFileToDelete picks which file to delete and returns the integer ID of this file.
+// Returns:
+//	- ID: returns the ID of the file to be deleted.
+func pickFileToDelete() int32 {
+	const (
+		beg                    = 10 // we can delete files staring from the file Hermes_10
```
alicjakwie: change beg to begin (more descriptive)
also since we changed our indexing (index from 0 but 0 is nil file) and we keep files 1-10 for persistence (right?) then we want to delete files from 11 to 50 included so begin = 11 and numberOfDeletableFiles = 40

alicjakwie (changes requested):
- [x] Change constants in PickFileToDelete and the name from beg to begin (left a comment)[https://github.com/googleinterns/step224-2020/pull/61/files#r489189507](url)

hermes/probe/delete/delete_file.go
evanSpendlove: Ack.

client/client_test.go
```diff
@@ -56,10 +54,10 @@ func setupCloudproberAndClient(t *testing.T) (context.Context, *hermes.Hermes, *
 		t.Fatalf("Cloudprober gRPC Client could not be initialised: %v", err)
 	}
 
-	return context.Background(), testHermes, client
+	return context.Background(), client
```
cianbgoog: What do we use this context for? Could we just not return it?

hermes/probe/delete/delete_file.go
```diff
@@ -0,0 +1,143 @@
...
+//		-> logger can log to Google Cloud if configured to do so.
+// Returns:
+//	- fileID: returns the ID of the file delete OR a missing file to be created if one is found.
+//	- err:
+//		Status:
+//		- StateJournalInconsistent: the file to be deleted does not exist in Hermes' StateJournal.
+//		- FileMissing: the file to be deleted could not be found in the target bucket.
+//		- BucketMissing: the target bucket on this target system was not found.
+//		- ProbeFailed: there was an error during one of the API calls and the probe failed.
+func DeleteRandomFile(ctx context.Context, target *probe.Target, client stiface.Client, logger *logger.Logger) (int32, error) {
```
cianbgoog: Do we need a function with 5 parameters, and 2 return values, that's only two lines? Or could we just call this directly?

hermes/probe/delete/delete_file.go
```diff
@@ -0,0 +1,143 @@
...
+	fileID := pickFileToDelete()
+	return DeleteFile(ctx, fileID, target, client, logger)
+}
+
+// DeleteFile deletes the file, corresponding to the ID passed, in the target storage system bucket.
+// It then checks that the file has been deleted by trying to get the object.
+// Arguments:
+//	- ctx: pass the context so this probe can be cancelled if needed.
+//	- config: pass the HermesProbeDef config for the probe calling this function.
```
cianbgoog: I suspect you could remove the `pass the` from all these comments, and the meaning would be equally useful. In many cases, there's not actually much useful data with some of these - e.g. most go developers will know what `ctx` means, and if you don't already know what `target` is, your comment leaves you none the wiser :-)

hermes/probe/delete/delete_file_test.go
```diff
@@ -0,0 +1,178 @@
...
+		Journal: &journalpb.StateJournal{
+			Intent:    &journalpb.Intent{},
+			Filenames: filenames,
+		},
+		LatencyMetrics: metrics,
+	}
+}
+
+// setupMockSystem sets up the mock storage system through the fake client.
+func setupMockSystem(ctx context.Context, t *testing.T) stiface.Client {
```
cianbgoog: s/Mock/Fake/, here and elsewhere.

cianbgoog (changes requested):

client/client_test.go
evanSpendlove: The context is used in all of the client method calls, but I've refactored this so that we create the context in the test and pass it to this method, rather than receiving it as a return value.

hermes/probe/delete/delete_file.go
evanSpendlove: This may be a slight over-complication... I'll remove this function and export `pickFileToDelete()` instead.

hermes/probe/delete/delete_file.go
evanSpendlove: I've removed `pass the` from the comments and tried to make them more descriptive. For the intuitive arguments, such as `ctx`, `target`, `logger`, would you recommend removing the line or just leaving it blank?

client/client_test.go
```diff
@@ -56,10 +54,10 @@ func setupCloudproberAndClient(t *testing.T) (context.Context, *hermes.Hermes, *
...
 }
 
-func teardownCloudproberAndClient(ctx context.Context, t *testing.T, testHermes *hermes.Hermes, client *CloudproberClient) {
+func teardownCloudproberAndClient(ctx context.Context, t *testing.T, client *CloudproberClient) {
```
leczb: This is for line 64.
Nit: Use short variable names for small-scoped variables. Rename `probesList` to `probes`.
Rename the `probe` variable on line 69 to `p`.

leczb: This is for line 66:
Nit: For set-up failures, just log the error as is. Only use the "got ...; want ..." format for assertions.

leczb: This is for line 78:
The first line of the comment should start with the identifier.
Move TODO entries to the end of the GoDoc comment.

hermes/probe/delete/delete_file_test.go
evanSpendlove: Thank you, I've changed this as requested.

client/client_test.go
```diff
@@ -97,8 +95,8 @@ func generateRedisProbeDef(name string) *probes_configpb.ProbeDef {
 
 // TestRegisterAndAddProbe tests that an extension probe type can be registered and added to Cloudprober without error.
 func TestRegisterAndAddProbe(t *testing.T) {
-	ctx, testHermes, client := setupCloudproberAndClient(t)
-	defer teardownCloudproberAndClient(ctx, t, testHermes, client)
+	ctx, client := setupCloudproberAndClient(t)
```
leczb: No need for `ctx` here. Just use context.Background() everywhere.

hermes/probe/delete/delete_file.go
```diff
@@ -0,0 +1,143 @@
...
+	"math/rand"
+	"time"
+
+	"cloud.google.com/go/storage"
+	"github.com/google/cloudprober/logger"
+	"github.com/googleapis/google-cloud-go-testing/storage/stiface"
+	"google.golang.org/api/iterator"
+
+	probe "github.com/googleinterns/step224-2020/hermes/probe"
+	m "github.com/googleinterns/step224-2020/hermes/probe/metrics"
```
leczb: Don't rename `metrics`.

hermes/probe/delete/delete_file.go
```diff
@@ -0,0 +1,143 @@
...
+	pb "github.com/googleinterns/step224-2020/hermes/proto"
+)
+
+// DeleteRandomFile deletes a random file in the target storage system bucket.
```
leczb: I would clarify that it deletes a random Hermes file that matches the file name pattern...
Otherwise the reader might think picks *any* random file in the bucket.

hermes/probe/delete/delete_file.go
```diff
@@ -0,0 +1,143 @@
...
+// Returns:
+//	- fileID: returns the ID of the file delete OR a missing file to be created if one is found.
+//	- err:
+//		Status:
+//		- StateJournalInconsistent: the file to be deleted does not exist in Hermes' StateJournal.
+//		- FileMissing: the file to be deleted could not be found in the target bucket.
+//		- BucketMissing: the target bucket on this target system was not found.
+//		- ProbeFailed: there was an error during one of the API calls and the probe failed.
+func DeleteRandomFile(ctx context.Context, target *probe.Target, client stiface.Client, logger *logger.Logger) (int32, error) {
+	fileID := pickFileToDelete()
```
leczb: No need for the intermediate variable.

hermes/probe/delete/delete_file.go
```diff
@@ -0,0 +1,143 @@
...
+	return DeleteFile(ctx, fileID, target, client, logger)
+}
+
+// DeleteFile deletes the file, corresponding to the ID passed, in the target storage system bucket.
```
leczb: Does this have to be exported?

hermes/probe/delete/delete_file.go
```diff
@@ -0,0 +1,143 @@
...
+// It then checks that the file has been deleted by trying to get the object.
+// Arguments:
+//	- ctx: pass the context so this probe can be cancelled if needed.
+//	- config: pass the HermesProbeDef config for the probe calling this function.
+//	- target: pass the target run information.
+//	- client: pass an initialised storage client for this target system.
+//	- logger: pass the logger associated with the probe calling this function.
+// Returns:
+//	- fileID: returns the ID of the file delete OR a missing file to be created if one is found.
```
leczb: Check the grammar.

hermes/probe/delete/delete_file.go
```diff
@@ -0,0 +1,143 @@
...
+//		- FileMissing: the file to be deleted could not be found in the target bucket.
+//		- BucketMissing: the target bucket on this target system was not found.
+//		- ProbeFailed: there was an error during one of the API calls and the probe failed.
+func DeleteFile(ctx context.Context, fileID int32, target *probe.Target, client stiface.Client, logger *logger.Logger) (int32, error) {
+	bucket := target.Target.GetBucketName()
+
+	// TODO(evanSpendlove): Add custom error object to return value and modify all returns.
+	filename, ok := target.Journal.Filenames[fileID]
+	if !ok {
+		return fileID, fmt.Errorf("delete(%q, %q) failed; status StateJournalInconsistent: expected fileID %d to exist", bucket, filename, fileID)
```
leczb: Suggested wording:

fmt.Errorf("StateJournalInconsistent: Journal.Filenames has no entry with file ID = %d", fileID)

Note that `filename` will be set to the empty string if the key lookup fails, so there's no point in including the value of `filename` in the error message.

hermes/probe/delete/delete_file.go
```diff
@@ -0,0 +1,143 @@
...
+			return fileID, fmt.Errorf("delete(%q, %q) failed; status %v: object %v still listed after delete", bucket, filename, status, obj.Name)
+		}
+		if err != nil {
+			status := m.BucketMissing
+			return fileID, fmt.Errorf("delete(%q, %q) failed; status %v: %w", bucket, filename, status, err)
+		}
+	}
+
+	// Update in-memory NIL file after delete operation.
+	delete(target.Journal.Filenames, int32(fileID))
```
leczb: No need for the type conversion: `fileID` is already `int32`.

hermes/probe/delete/delete_file.go
```diff
@@ -0,0 +1,143 @@
...
+// pickFileToDelete picks which file to delete and returns the integer ID of this file.
+// Returns:
+//	- ID: returns the ID of the file to be deleted.
+func pickFileToDelete() int32 {
+	const (
+		begin                  = 11 // we can delete files staring from the file Hermes_11
+		numberOfDeletableFiles = 40 // there are 40 files to delete from [Hermes_11,Hermes_50]
+	)
+	rand.Seed(time.Now().UnixNano())
+	return int32(rand.Intn(numberOfDeletableFiles) + being)
```
leczb: Fix typo.
Also, make sure code compiles before sending it for review.

hermes/probe/delete/delete_file.go
```diff
@@ -0,0 +1,143 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Delete_file implements the probe operation for deleting a file in a
+// storage system.
+
+// Package delete implements the file deletion operation with a storage system.
+package delete
```
leczb: Make the filename match the package name.

hermes/probe/delete/delete_file_test.go
```diff
@@ -0,0 +1,178 @@
...
+	client := mock.NewFakeClient()
+
+	mockBucket := client.Bucket(bucketName)
+	if err := mockBucket.Create(ctx, "", nil); err != nil {
+		t.Fatalf("could not create mock bucket when setting up mock storage system, err: %v", err)
```
leczb: No need to include "err :" in the error message. Just `failed to create bucket: %v"`.
Keep error messages succinct.

hermes/probe/delete/delete_file_test.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	for i := firstID; i <= lastID; i++ {
+		var id string
+		if i <= 9 {
+			id = fmt.Sprintf("0%d", i)
```
leczb: You can replace these two branches by specifying "%02d" in the format:
`filename := fmt.Sprintf("Hermes_%02d_%s", id, hash)`

hermes/probe/delete/delete_file_test.go
```diff
@@ -0,0 +1,178 @@
...
+		} else {
+			id = fmt.Sprintf("%d", i)
+		}
+		filename := fmt.Sprintf("Hermes_%s_%s", id, hash)
+
+		writer := client.Bucket(bucketName).Object(filename).NewWriter(ctx)
+		n, err := writer.Write([]byte(contents))
+		if err != nil || n != len([]byte(contents)) {
+			t.Fatalf("failed to create file during setup of mock storage system, err: %v", err)
```
leczb: `err` might be nil. I would split the two cases:
if err != nil { ...
}
if n != len(...) {
  t.Fatalf("short write: wrote %d bytes; wanted %d", n, len(...))
}

hermes/probe/delete/delete_file_test.go
```diff
@@ -0,0 +1,178 @@
...
+	target := genTestTarget(genTestConfig(testProbeName), t)
+
+	logger, err := logger.NewCloudproberLog(testProbeName)
+	if err != nil {
+		t.Fatalf("error in initializing logger for the probe (%s): %v", testProbeName, err)
+	}
+
+	fileID, err := DeleteRandomFile(ctx, target, client, logger)
+	if err != nil {
+		t.Errorf("deleteRandomFile(ID: %d) failed: expected error as %v, got %v", fileID, nil, err)
```
leczb: Nit: Fix the name in the error message: replace `deleteRandomFile` with `DeleteRandomFile`.

hermes/probe/delete/delete_file_test.go
```diff
@@ -0,0 +1,178 @@
...
+	for {
+		obj, err := objects.Next()
+		if err == iterator.Done {
+			break
+		}
+		if obj.Name == filename {
+			t.Errorf("deleteRandomFile failed, expected object to be deleted, got object found.")
+		}
+		if err != nil {
+			t.Errorf("deleteRandomFile failed, expected %v, got %v", nil, err)
```
leczb: Assuming this is an unexpected error, I would make this `t.Fatalf()`.

leczb (changes requested):

client/client_test.go
evanSpendlove: Ack.

evanSpendlove: Ack.

evanSpendlove: Ack.

client/client_test.go
evanSpendlove: Ack.

hermes/probe/delete/delete_file.go
evanSpendlove: Ack.

hermes/probe/delete/delete_file.go
evanSpendlove: I've removed this function in a later commit, so pickFileToDelete() is now exported as `PickFileToDelete()` and called in the probing algorithm.

hermes/probe/delete/delete_file.go
evanSpendlove: It did previously also have to be as it is used in check_nil.go, but now this is the only function so it is exported.

hermes/probe/delete/delete_file.go
evanSpendlove: Ack.

client/client_test.go
```diff
@@ -97,8 +95,8 @@ func generateRedisProbeDef(name string) *probes_configpb.ProbeDef {
...
+	defer teardownCloudproberAndClient(ctx, t, client)
 
 	for i := 0; i < testRunCount; i++ {
 		probeName := fmt.Sprintf("testExtension%d", i)
```
deniadelcid, line 103: Can we remove the use of Redis probes in this file and instead make use of the MonitorProbe or HermesProbeDef (whichever is a better fit)?

hermes/probe/delete/delete_file.go
evanSpendlove: Ack.

hermes/probe/delete/delete_file.go
evanSpendlove: I've changed this to the suggested wording. 

Okay, I'll make sure not to include the key if the lookup fails in future.

hermes/probe/delete/delete_file.go
evanSpendlove: Typo fixed, apologies.

hermes/probe/delete/delete_file.go
evanSpendlove: Thank you for pointing this out, I've amended it.

hermes/probe/delete/delete_file.go
evanSpendlove: I've made this change.

Also, go-lint has mentioned that `delete.DeleteFile()` stutters.
I feel that `delete.File()` is strange, but that might just be the go convention.

Do you have any suggestions about how I can alter this?

hermes/probe/delete/delete_file_test.go
evanSpendlove: Ack.

hermes/probe/delete/delete_file_test.go
evanSpendlove: That's very neat, thank you!

hermes/probe/delete/delete_file_test.go
evanSpendlove: Ack.

hermes/probe/delete/delete_file_test.go
evanSpendlove: DeleteRandomFile() has been removed.

hermes/probe/delete/delete_file_test.go
evanSpendlove: Okay, thank you.

client/client_test.go
```diff
@@ -95,17 +95,17 @@ func generateRedisProbeDef(name string) *probes_configpb.ProbeDef {
...
 func TestRegisterAndAddProbe(t *testing.T) {
-	ctx, client := setupCloudproberAndClient(t)
-	defer teardownCloudproberAndClient(ctx, t, client)
+	client := setupCloudproberAndClient(t)
+	defer teardownCloudproberAndClient(t, client)
 
 	for i := 0; i < testRunCount; i++ {
 		probeName := fmt.Sprintf("testExtension%d", i)
-		if err := client.RegisterAndAddProbe(ctx, int(myprobe.E_RedisProbe.Field), generateRedisProbeDef(probeName), &myprobe.Probe{}); err != nil {
+		if err := client.RegisterAndAddProbe(context.Background(), int(myprobe.E_RedisProbe.Field), generateRedisProbeDef(probeName), &myprobe.Probe{}); err != nil {
```
cianbgoog: Just create a context.Background at the start, rather than constantly repeating it. Here and elsewhere.

cianbgoog (approved):
happy to merge if you create one `context.Context` per test, rather than constantly using `context.Background()`

client/client_test.go
evanSpendlove: I changed it to calling context.Background() as per Balazs last comment:
    "No need for ctx here. Just use context.Background() everywhere."

Would you recommend `ctx := context.Background()` as per the previous version?

leczb: Sorry, I should have stated it better: there's no reason to return a Context from the setup function, as context.Background() is to be used in tests by default (unless something else is required).
It's OK to declare a `ctx` variable and set it to context.Backround() if there are many calls in the test that expect a context. It's also OK to just use context.Background(). Your choice.

leczb: Why are we using a package from the "examples" dir?

hermes/probe/delete/delete.go
```diff
@@ -13,9 +13,9 @@
 // limitations under the License.
 //
 // Delete_file implements the probe operation for deleting a file in a
-// storage system.
+// storage systemetrics.
```
leczb: Typo.

hermes/probe/delete/delete.go
```diff
@@ -13,9 +13,9 @@
...
 
-// Package delete implements the file deletion operation with a storage system.
+// Package delete implements the file deletion operation with a storage systemetrics.
```
leczb: Typo.

hermes/probe/delete/delete.go
```diff
@@ -89,55 +73,55 @@ func DeleteFile(ctx context.Context, fileID int32, target *probe.Target, client
...
 		case storage.ErrBucketNotExist:
-			status = m.BucketMissing
+			status = metrics.BucketMissing
 		default:
-			status = m.ProbeFailed
+			status = metrics.ProbeFailed
 		}
 
-		target.LatencyMetrics.APICallLatency[m.APIDeleteFile][status].Metric("hermes_api_latency_s").AddFloat64(time.Now().Sub(start).Seconds())
+		target.LatencyMetrics.APICallLatency[metrics.APIDeleteFile][status].Metric("hermes_api_latency_seconds").AddFloat64(time.Now().Sub(start).Seconds())
```
leczb: "hermes_api_latency_seconds" is repeated multiple times. It should be a string constant.

hermes/probe/delete/delete.go
```diff
@@ -89,55 +73,55 @@ func DeleteFile(ctx context.Context, fileID int32, target *probe.Target, client
...
 }
 
-// pickFileToDelete picks which file to delete and returns the integer ID of this file.
+// PickFileToDelete picks which file to delete and returns the integer ID of this file.
 // Returns:
 //	- ID: returns the ID of the file to be deleted.
-func pickFileToDelete() int32 {
+func PickFileToDelete() int32 {
 	const (
 		begin                  = 11 // we can delete files staring from the file Hermes_11
```
leczb: Typo: *starting.

hermes/probe/delete/delete_test.go
```diff
@@ -114,28 +114,25 @@ func genTestTarget(cfg *monitorpb.HermesProbeDef, t *testing.T) *probe.Target {
...
-// setupMockSystem sets up the mock storage system through the fake client.
-func setupMockSystem(ctx context.Context, t *testing.T) stiface.Client {
+// setupFakeSystem sets up the fake storage system through the fake client.
+func setupFakeSystem(ctx context.Context, t *testing.T) stiface.Client {
 	client := mock.NewFakeClient()
 
 	mockBucket := client.Bucket(bucketName)
 	if err := mockBucket.Create(ctx, "", nil); err != nil {
-		t.Fatalf("could not create mock bucket when setting up mock storage system, err: %v", err)
+		t.Fatalf("failed to create mock bucket, err: %v", err)
```
leczb: It's not a mock.

hermes/probe/delete/delete_test.go
```diff
@@ -114,28 +114,25 @@ func genTestTarget(cfg *monitorpb.HermesProbeDef, t *testing.T) *probe.Target {
...
 
 		writer := client.Bucket(bucketName).Object(filename).NewWriter(ctx)
 		n, err := writer.Write([]byte(contents))
-		if err != nil || n != len([]byte(contents)) {
-			t.Fatalf("failed to create file during setup of mock storage system, err: %v", err)
+		if err != nil {
+			t.Fatalf("failed to create file, : %v", err)
+		}
+		if n != len([]byte(contents)) {
+			t.Fatalf("short write: wrote %d butes; wanted %d", n, len([]byte(contents)))
```
leczb: Typo: *bytes.

hermes/probe/delete/delete_test.go
```diff
@@ -114,28 +114,25 @@ func genTestTarget(cfg *monitorpb.HermesProbeDef, t *testing.T) *probe.Target {
 	}
 }
 
-// setupMockSystem sets up the mock storage system through the fake client.
-func setupMockSystem(ctx context.Context, t *testing.T) stiface.Client {
+// setupFakeSystem sets up the fake storage system through the fake client.
+func setupFakeSystem(ctx context.Context, t *testing.T) stiface.Client {
```
leczb: Nit: I would call this `fakeStorageClient`.
Then the call-site reads better:
```
client := fakeStorageClient(...)
```

hermes/probe/delete/delete_test.go
```diff
@@ -114,28 +114,25 @@ func genTestTarget(cfg *monitorpb.HermesProbeDef, t *testing.T) *probe.Target {
...
 
 	for i := firstID; i <= lastID; i++ {
-		var id string
-		if i <= 9 {
-			id = fmt.Sprintf("0%d", i)
-		} else {
-			id = fmt.Sprintf("%d", i)
-		}
-		filename := fmt.Sprintf("Hermes_%s_%s", id, hash)
+		filename := fmt.Sprintf("Hermes_%02d_%s", i, hash)
```
leczb, line 126: Where is `hash` coming from?

hermes/probe/delete/delete_test.go
```diff
@@ -170,7 +167,7 @@ func TestDeleteRandomFile(t *testing.T) {
 			t.Errorf("deleteRandomFile failed, expected object to be deleted, got object found.")
 		}
 		if err != nil {
-			t.Errorf("deleteRandomFile failed, expected %v, got %v", nil, err)
+			t.Fatalf("deleteRandomFile failed, expected %v, got %v", nil, err)
```
leczb: There's no expectation here, so the "expected" part is not required.
Use this message format for these kinds of failures `t.Fatalf("<operation desc> failed: %v", err)`.

leczb (changes requested):

client/client_test.go
evanSpendlove: Ah okay, I misunderstood, apologies. I've amended this to use a single `ctx` variable per test.

hermes/probe/delete/delete.go
evanSpendlove: Refactoring... Thanks for spotting this, I've amended it.

hermes/probe/delete/delete.go
evanSpendlove: Ack.

hermes/probe/delete/delete.go
evanSpendlove: Ack.

hermes/probe/delete/delete_test.go
evanSpendlove: Ack.

hermes/probe/delete/delete_test.go
evanSpendlove: There's a static hash calculated and stored in a global variable, as the file contents are the same for each file.

hermes/probe/delete/delete.go
```diff
@@ -0,0 +1,127 @@
...
+//	- err:
+//		Status:
+//		- StateJournalInconsistent: the file to be deleted does not exist in Hermes' StateJournal.
+//		- FileMissing: the file to be deleted could not be found in the target bucket.
+//		- BucketMissing: the target bucket on this target system was not found.
+//		- ProbeFailed: there was an error during one of the API calls and the probe failed.
+func DeleteFile(ctx context.Context, fileID int32, target *probe.Target, client stiface.Client, logger *logger.Logger) (int32, error) {
+	bucket := target.Target.GetBucketName()
+
+	if fileID <= 10 || fileID >= 51 {
```
deniadelcid: Don't hard code these digits, use constants or flags instead.

hermes/probe/delete/delete.go
```diff
@@ -0,0 +1,127 @@
...
+		return fileID, fmt.Errorf("delete(%q, %q) failed; status invalidArgument: expected fileID %d to be within valid inclusive range: 11-50", bucket, fileID, fileID)
```
deniadelcid: Nit: make your error prefixes match the function the function name where they actually failed (i.e. s/delete/DeleteFile). This makes debugging easier.

deniadelcid: Also, instead of hardcoding the status in the error string, could you add support for statuses and use the ones that you've created in metrics.go instead?

If there isn't an applicable error there (e..g. OutOfRange or InvalidArgument), feel free to add it to metrics.go we just open an issue to add the error type to metrics.go in a future PR.

client/client_test.go
evanSpendlove: At the time of writing this test, we hadn't started writing an extension probe. This example provides a working extension probe implementation. There is a TODO to change this to our probe implementation when it is ready.

hermes/probe/delete/delete_test.go
evanSpendlove: Okay, that's a good point. I've made this change.

hermes/probe/delete/delete.go
evanSpendlove: I've refactored this into a constant and used it throughout.

I've also added a TODO to add this into metrics.go

hermes/probe/delete/delete_test.go
evanSpendlove: I've changed this to fake.

hermes/probe/delete/delete_test.go
evanSpendlove: Okay, thanks. I've made this change.

client/client_test.go
```diff
@@ -60,14 +60,15 @@ func setupCloudproberAndClient(t *testing.T) *CloudproberClient {
...
-	probes, err := client.ListProbes(context.Background())
+	probes, err := client.ListProbes(ctx)
 	if err != nil {
 		t.Fatalf("ListProbes() failed during teardown: %v", err)
 	}
 
 	for _, p := range probes {
-		if err = client.RemoveProbe(context.Background(), p.GetName()); err != nil {
+		if err = client.RemoveProbe(ctx, p.GetName()); err != nil {
 			t.Errorf("RemoveProbe() failed, expected error: %v, got %v", nil, err)
```
leczb: Fatalf
Also, simplify the error message, as this is not a test expectation:

t.Fatalf("RemoveProbe failed: %v", err)

client/client_test.go
```diff
@@ -95,17 +96,18 @@ func generateRedisProbeDef(name string) *probes_configpb.ProbeDef {
...
 func TestRegisterAndAddProbe(t *testing.T) {
+	ctx := context.Background()
 	client := setupCloudproberAndClient(t)
 	defer teardownCloudproberAndClient(t, client)
 
 	for i := 0; i < testRunCount; i++ {
 		probeName := fmt.Sprintf("testExtension%d", i)
-		if err := client.RegisterAndAddProbe(context.Background(), int(myprobe.E_RedisProbe.Field), generateRedisProbeDef(probeName), &myprobe.Probe{}); err != nil {
+		if err := client.RegisterAndAddProbe(ctx, int(myprobe.E_RedisProbe.Field), generateRedisProbeDef(probeName), &myprobe.Probe{}); err != nil {
 			t.Errorf("Probe not correctly registered and added to Cloudprober, expected error: %v, got: %v", nil, err)
```
leczb: No need for a formatting verb if the value is nil.
Simplify the error message like
```
t.Errorf("RegisterAndAddProbe failed: %v", err)
```

client/client_test.go
```diff
@@ -95,17 +96,18 @@ func generateRedisProbeDef(name string) *probes_configpb.ProbeDef {
...
 		}
 
 		// Check if the probe was added correctly
-		probesList, err := client.ListProbes(context.Background())
+		probesList, err := client.ListProbes(ctx)
 		if err != nil {
 			t.Errorf("ListProbes() failed, expected error: %v, got: %v", nil, err)
```
leczb: `t.Errorf("ListProbes failed: %v")`

client/client_test.go
```diff
@@ -118,17 +120,18 @@ func TestRegisterAndAddProbe(t *testing.T) {
 
 // TestRemoveProbe tests that the RemoveProbe() method removes a probe from Cloudprober
```
leczb: Nit: there's no need to comment test functions that have a self-describing name. The assumption on the reader's side is that `TestFoo()` tests `Foo()`.

client/client_test.go
```diff
@@ -118,17 +120,18 @@ func TestRegisterAndAddProbe(t *testing.T) {
...
 	for i := 0; i < testRunCount; i++ {
 		probeName := fmt.Sprintf("testExtension%d", i)
-		if err := client.RegisterAndAddProbe(context.Background(), int(myprobe.E_RedisProbe.Field), generateRedisProbeDef(probeName), &myprobe.Probe{}); err != nil {
+		if err := client.RegisterAndAddProbe(ctx, int(myprobe.E_RedisProbe.Field), generateRedisProbeDef(probeName), &myprobe.Probe{}); err != nil {
 			t.Errorf("Probe not correctly registered and added to Cloudprober, expected error: %v, got: %v", nil, err)
 		}
 
 		// Check if the probe was added correctly
-		probesList, err := client.ListProbes(context.Background())
+		probesList, err := client.ListProbes(ctx)
```
leczb: Simplify the variable name: `probes`.

client/client_test.go
```diff
@@ -118,17 +120,18 @@ func TestRegisterAndAddProbe(t *testing.T) {
...
 		if err != nil {
 			t.Errorf("ListProbes() failed, expected error: %v, got: %v", nil, err)
```
leczb: This should be `Fatalf`.

client/client_test.go
```diff
@@ -155,17 +158,18 @@ func TestRemoveProbe(t *testing.T) {
...
 func TestListProbes(t *testing.T) {
+	ctx := context.Background()
 	client := setupCloudproberAndClient(t)
 	defer teardownCloudproberAndClient(t, client)
 
 	for i := 0; i < testRunCount; i++ {
 		probeName := fmt.Sprintf("testExtension%d", i)
-		if err := client.RegisterAndAddProbe(context.Background(), int(myprobe.E_RedisProbe.Field), generateRedisProbeDef(probeName), &myprobe.Probe{}); err != nil {
+		if err := client.RegisterAndAddProbe(ctx, int(myprobe.E_RedisProbe.Field), generateRedisProbeDef(probeName), &myprobe.Probe{}); err != nil {
 			t.Errorf("Probe not correctly registered and added to Cloudprober, expected error: %v, got: %v", nil, err)
```
leczb: Replace Errorf with Fatalf.
Rationale: If RegisterAndAddProbe fails, it means there's a problem with the environment or the test code itself. There's no point in continuing the test.
In general test set-up/tear-down should use Fatalf. Check and fix this throughout the codebase.

hermes/probe/delete/delete.go
```diff
@@ -13,9 +13,9 @@
 // limitations under the License.
 //
 // Delete_file implements the probe operation for deleting a file in a
-// storage systemetrics.
+// storage system.
 
-// Package delete implements the file deletion operation with a storage systemetrics.
+// Package delete implements the file deletion operation with a storage system.
```
leczb, line 18: Please verify that GoDoc do the right thing and renders line 18 as the package doc.

hermes/probe/delete/delete.go
```diff
@@ -55,7 +59,7 @@ func DeleteFile(ctx context.Context, fileID int32, target *probe.Target, client
 	bucket := target.Target.GetBucketName()
 
 	if fileID <= 10 || fileID >= 51 {
-		return fileID, fmt.Errorf("delete(%q, %q) failed; status invalidArgument: expected fileID %d to be within valid inclusive range: 11-50", bucket, fileID, fileID)
+		return fileID, fmt.Errorf("delete(%q, %q) failed; invalidArgument: expected fileID %d to be within valid inclusive range: 11-50", bucket, fileID, fileID)
```
leczb: Nit: Replace "invalidArgument" with "invalid argument". "invalidArgument" suggests that it's some kind of symbol, which it isn't.

hermes/probe/delete/delete.go
```diff
@@ -83,29 +87,31 @@ func DeleteFile(ctx context.Context, fileID int32, target *probe.Target, client
 			status = metrics.ProbeFailed
 		}
 
-		target.LatencyMetrics.APICallLatency[metrics.APIDeleteFile][status].Metric("hermes_api_latency_seconds").AddFloat64(time.Now().Sub(start).Seconds())
+		target.LatencyMetrics.APICallLatency[metrics.APIDeleteFile][status].Metric(apiLatency).AddFloat64(time.Now().Sub(start).Seconds())
```
leczb, line 93: This update should be factored out as it repeats many times throughout the code.

hermes/probe/delete/delete_test.go
```diff
@@ -101,7 +101,7 @@ func genTestTarget(cfg *monitorpb.HermesProbeDef, t *testing.T) *probe.Target {
 
 	metrics, err := m.NewMetrics(cfg, genTargetPb())
 	if err != nil {
-		t.Fatalf("could not initialise metrics using config and target provided, err: %v", err)
+		t.Fatalf("could not initialise metrics using config and target provided, %v", err)
```
leczb: Nit: Use ": " as the separator instead of ", ".

hermes/probe/delete/delete_test.go
```diff
@@ -114,13 +114,14 @@ func genTestTarget(cfg *monitorpb.HermesProbeDef, t *testing.T) *probe.Target {
 	}
 }
 
-// setupFakeSystem sets up the fake storage system through the fake client.
-func setupFakeSystem(ctx context.Context, t *testing.T) stiface.Client {
+// fakeStorageClient sets up the fake storage system through the fake client
+// and returns the fake client.
+func fakeStorageClient(ctx context.Context, t *testing.T) stiface.Client {
 	client := mock.NewFakeClient()
```
leczb: Rename "mock" to "fakeclient", and "NewFakeClient" to "New".
Then this becomes `client := fakeclient.New()`, which is more idiomatic.

hermes/probe/delete/delete_test.go
```diff
@@ -114,13 +114,14 @@ func genTestTarget(cfg *monitorpb.HermesProbeDef, t *testing.T) *probe.Target {
...
 
-	mockBucket := client.Bucket(bucketName)
-	if err := mockBucket.Create(ctx, "", nil); err != nil {
-		t.Fatalf("failed to create mock bucket, err: %v", err)
+	fakeBucket := client.Bucket(bucketName)
```
leczb: Just `bucket`. The reader already knows everything is a fake here.

hermes/probe/delete/delete_test.go
```diff
@@ -114,13 +114,14 @@ func genTestTarget(cfg *monitorpb.HermesProbeDef, t *testing.T) *probe.Target {
 	}
 }
 
-// setupFakeSystem sets up the fake storage system through the fake client.
-func setupFakeSystem(ctx context.Context, t *testing.T) stiface.Client {
+// fakeStorageClient sets up the fake storage system through the fake client
```
leczb: Nit: This is not technically correct. The fake client implements a fake client, not a fake storage system (which is the server-side).
If the fake implemented the storage system, we could use a real client (this would actually be preferred).

hermes/probe/delete/delete_test.go
```diff
@@ -144,7 +145,7 @@ func TestDeleteRandomFile(t *testing.T) {
 	testProbeName := "testDelete1"
 	ctx := context.Background()
 
-	client := setupFakeSystem(ctx, t)
+	client := fakeStorageClient(ctx, t)
```
leczb: Optional: I think it would be cleaner if `fakeStorageClient()` was called `populateBucket` or `createTestFiles` (or something to that effect) and it only did the population. Then the test code would look like this:

```
client := fakeclient.New()
if err := createTestFiles(t, client); err != nil {
   ...
}
```

leczb (changes requested):

hermes/probe/delete/delete.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// PickFileToDelete picks which file to delete and returns the integer ID of this file.
+// Returns:
+//	- ID: returns the ID of the file to be deleted.
+func PickFileToDelete() int32 {
+	const (
+		begin                  = 11 // we can delete files starting from the file Hermes_11
+		numberOfDeletableFiles = 40 // there are 40 files to delete from [Hermes_11,Hermes_50]
+	)
```
deniadelcid: Constants should go towards the top of the file, right below your imports

cianbgoog (approved):
Approved, subject to you factoring out the metrics update as leczb asks.

client/client_test.go
evanSpendlove: I've changed this accordingly.

client/client_test.go
evanSpendlove: I've changed this accordingly.

client/client_test.go
evanSpendlove: Okay, I've changed this accordingly.

client/client_test.go
evanSpendlove: I've changed this throughout.

client/client_test.go
evanSpendlove: Ack.

hermes/probe/delete/delete.go
evanSpendlove: Okay, changed.

hermes/probe/delete/delete_test.go
evanSpendlove: Ack.

hermes/probe/delete/delete_test.go
evanSpendlove: Ack.

hermes/probe/delete/delete_test.go
evanSpendlove: I've changed this to `createTestFiles()` and updated the mock package I'm using to the latest in Alicja's PR, so it's now `fakegcs`.

hermes/probe/delete/delete_test.go
evanSpendlove: I've updated to the latest commit in Alicja's PR for fakes, so it's now:
`fakegcs.NewClient()` which is quite intuitive, I think.

hermes/probe/delete/delete_test.go
evanSpendlove: Ah okay, I've removed this comment.

client/client_test.go
evanSpendlove: I disagree, `ListProbes()` is one of the methods under test here, and it is useful to see the results of the rest of the test rather than exiting here. For example, if `ListProbes()` fails in one specific test multiple times, but not in the others, that would be useful to know.

client/client_test.go
evanSpendlove: While `RegisterAndAddProbe()` is a setup for testing `RemoveProbe()` and `ListProbes()`, it is still one of the methods under test in this package, and it is very useful to not exit the test the first time that it fails. 

I would prefer to keep the `Errorf()` for calls to add, remove and list probes. It can be helpful to see if a method fails throughout the remainder of the test, just because it failed once.

hermes/probe/delete/delete.go
evanSpendlove: I'm currently unable to run GoDoc on my code (pinged you with explanation).

hermes/probe/delete/delete_test.go
```diff
@@ -0,0 +1,176 @@
...
+		Name:                   "hermes",
+		TargetSystem:           monitorpb.Target_GOOGLE_CLOUD_STORAGE,
+		TotalSpaceAllocatedMib: int64(100),
+		BucketName:             bucketName,
+	}
+}
+
+const (
+	firstID    = int32(1)
+	lastID     = int32(50)
```
deniadelcid, line 83: Can you add a TODO wherever it's relevant to change the datatypes of file IDs to `int`? You can reference issue [#76 ](https://github.com/googleinterns/step224-2020/issues/76) in the comment.

deniadelcid: Just to be clear, leave these as they are for now. In some future PR we can update all of them at once.

hermes/probe/delete/delete.go
deniadelcid: I think this is also applicable on other files too, so I've opened issue #77 so this refactoring is done consistently throughout all files (it might also be necessary to tweak) metrics.go (?) in the process so lets do this in a different PR.

@evanSpendlove In the meantime, could you please add a TODO comment where relevant and reference #77?

hermes/probe/delete/delete.go
```diff
@@ -0,0 +1,133 @@
...
+	logger.Infof("Object %v deleted in bucket %s.", file, bucket)
+	return fileID, nil
+}
+
+// PickFileToDelete picks which file to delete and returns the integer ID of this file.
+// Returns:
+//	- ID: returns the ID of the file to be deleted.
+func PickFileToDelete() int32 {
+	const (
+		begin                  = 11 // we can delete files starting from the file Hermes_11
```
deniadelcid: Maybe rename to `minFileIDToDelete`? Also, I thought that only files 1-9 were idempotent.

hermes/probe/delete/delete_test.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	filename := target.Journal.Filenames[fileID]
+	objects := client.Bucket(bucketName).Objects(ctx, &storage.Query{Prefix: filename})
+	for {
+		obj, err := objects.Next()
+		if err == iterator.Done {
+			break
+		}
+		if obj.Name == filename {
+			t.Errorf("deleteRandomFile failed, expected object to be deleted, got object found.")
```
deniadelcid: nit
```suggestion
			t.Errorf("deleteRandomFile failed: expected object to be deleted, got object found.")
```

client/client_test.go
evanSpendlove: Not yet, it's noted as a TODO. It can be added when MonitorProbe is runnable.

hermes/probe/delete/delete_test.go
evanSpendlove: I'll add this as a TODO.

hermes/probe/delete/delete.go
evanSpendlove: I'll rename it accordingly. The nil file is file 0, so the first 10 files (1-10) are idempotent.

hermes/probe/delete/delete.go
evanSpendlove: Ack.

hermes/probe/delete/delete.go
evanSpendlove: Moved to constants.

hermes/probe/delete/delete.go
evanSpendlove: I've added a TODO.

hermes/probe/delete/delete_test.go
evanSpendlove: Ack.

hermes/probe/delete/delete.go
evanSpendlove: Yep, I'll add invalid argument as a status and open the issue for it.

hermes/probe/delete/delete.go
evanSpendlove: Ran GoDoc on my personal laptop.

It renders this line as the package doc as expected.

hermes/probe/delete/delete.go
```diff
@@ -35,7 +35,9 @@ import (
 )
 
 const (
-	apiLatency = "hermes_api_latency_seconds" // TODO(evanSpendlove) add this constant to metrics.go
+	apiLatency             = "hermes_api_latency_seconds" // TODO(evanSpendlove) add this constant to metrics.go
+	minFileIDToDelete      = 11                           // we can delete files starting from the file Hermes_11
+	numberOfDeletableFiles = 40                           // there are 40 files to delete from [Hermes_11,Hermes_50]
```
leczb: Nit: why not maxFileIDToDelete for symmetry?

hermes/probe/delete/delete.go
```diff
@@ -88,11 +91,12 @@ func DeleteFile(ctx context.Context, fileID int32, target *probe.Target, client
 		}
 
 		target.LatencyMetrics.APICallLatency[metrics.APIDeleteFile][status].Metric(apiLatency).AddFloat64(time.Now().Sub(start).Seconds())
-		return fileID, fmt.Errorf("delete(%q, %q) failed; status %v: %w", bucket, filename, status, err)
+		return fileID, fmt.Errorf("deleteFile(%q, %q) failed; status %v: %w", bucket, filename, status, err)
```
leczb: Nit: make it match the name of the function: s/deleteFile/DeleteFile/.
Error messages should start with lower-case, except when the first word is a proper name / identifier.
Here and elsewhere.

leczb (approved):

cianbgoog (approved):

hermes/probe/delete/delete.go
evanSpendlove: Ah okay, I didn't realise that identifiers over-ruled the lower-case rule, thanks.

I'll amend this.

hermes/probe/delete/delete.go
evanSpendlove: That's a good idea, I've changed it accordingly.

hermes/probe/delete/delete.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/googleinterns/step224-2020/hermes/probe/metrics"
+	"google.golang.org/api/iterator"
+
+	pb "github.com/googleinterns/step224-2020/hermes/proto"
+)
+
+const (
+	apiLatency             = "hermes_api_latency_seconds" // TODO(evanSpendlove) add this constant to metrics.go
+	minFileIDToDelete      = 11                           // we can delete files starting from the file Hermes_11
+	numberOfDeletableFiles = 40                           // there are 40 files to delete from [Hermes_11,Hermes_50]
```
deniadelcid: Just in case you missed leczb's comment. ;)
```suggestion
	maxFileIDToDelete = 40                           // there are 40 files to delete from [Hermes_11,Hermes_50]
```

deniadelcid (approved):

hermes/probe/delete/delete.go
evanSpendlove: Thanks :) I've made this change in a commit.

PR #63, opened 2020-09-15
Added ProbeError type for use within Hermes.

evanSpendlove:


hermes/probe/error/error.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"fmt"
+
+	"github.com/googleinterns/step224-2020/hermes/probe/metrics"
+)
+
+// ProbeError is an error that includes an exit status for used within Hermes.
+type ProbeError struct {
+	Status metrics.ExitStatus
```
leczb: Nit: "Exit status" has a particular meaning: it's the numeric result code returned by an exiting process on Unix/Linux systems. Using the "ExitStatus" name suggests to the reader that there's a subprocess execution happening somewhere and we record the Unix exit status from the subprocess.
I would rename `ExitStatus` to `APICallStatus` or something similar to avoid this confusion.
It's not really important, so I'll leave it to you whether you do the rename.

hermes/probe/error/error.go
```diff
@@ -0,0 +1,70 @@
...
+// Arguments:
+//	- status: pass the exit status associated with this error.
+//	- msg: pass the additional context of the error as a string.
+//	- innner: pass the inner error to be wrapped.
+// Returns:
+//	- ProbeError: returns a new ProbeError object containing the args passed.
+func Wrap(status metrics.ExitStatus, msg string, inner error) ProbeError {
+	return &ProbeError{
+		Status: status,
+		Err:    fmt.Errorf("%s: %w", msg, inner),
```
leczb: This is not wrapping, as it's transforming the underlying error, losing its type information.
Wrapping means we retain the original value of `inner`, so it can be unwrapped.
Wrapping is lossless.

hermes/probe/error/error.go
```diff
@@ -0,0 +1,70 @@
...
+
+// Wrap returns a new ProbeError containing the error passed
+// with additional context.and the status.
+// Arguments:
+//	- status: pass the exit status associated with this error.
+//	- msg: pass the additional context of the error as a string.
+//	- innner: pass the inner error to be wrapped.
+// Returns:
+//	- ProbeError: returns a new ProbeError object containing the args passed.
+func Wrap(status metrics.ExitStatus, msg string, inner error) ProbeError {
```
leczb: I would simplify the signature of this function to

`func Wrap(status metrics.ExitStatus, msg string)` or
`func Wrap(status metrics.ExitStatus, err error)`.

Rationale: the `Err` field is set to a new error value generated using fmt.Errorf(). This might as well happen at the call-site.

hermes/probe/error/error.go
```diff
@@ -0,0 +1,70 @@
...
+		Status: status,
+		Err:    fmt.Errorf("%s: %w", msg, inner),
+	}
+}
+
+// Error() returns the error string from the error.
+// Returns:
+//	- string: returns the error as a string.
+func (e *ProbeError) Error() string {
+	return fmt.Errorf("%v: %s: %w", e.Status, e.Msg, e.Inner)
```
leczb: This is not how it works.
fmt.Errorf() returns a struct that implements the `error` interface.
The `Error() string` method should return a string.

leczb (changes requested):

hermes/probe/error/error.go
cianbgoog: The %w verb appears to do magical wrapping with fmt.Errorf: https://blog.golang.org/go1.13-errors

evanSpendlove: Ah okay, I misunderstood the concept then. I've renamed this to `New()` as I think that is more appropriate.

hermes/probe/error/error.go
evanSpendlove: That makes sense, I've incorporated this change.

hermes/probe/error/error.go
evanSpendlove: Ah okay, I misunderstood this too. I've amended this.

hermes/probe/error/error.go
evanSpendlove: Okay, I wasn't aware of that. I've added it as a TODO, but I'd prefer not to make the change right now if possible.

hermes/probe/error/error.go
```diff
@@ -22,37 +22,38 @@
...
 // ProbeError is an error that includes an exit status for used within Hermes.
 type ProbeError struct {
+	// TODO(evanSpendlove): Refactor metrics.ExitStatus to use something similar to metrics.APICallStatus.
 	Status metrics.ExitStatus
 	Err    error
 }
 
-// Wrap returns a new ProbeError containing the error passed
+// New returns a new ProbeError containing the error passed
 // with additional context.and the status.
```
leczb: Nit: fix the comment.

hermes/probe/error/error.go
```diff
@@ -22,37 +22,38 @@
...
 // Arguments:
 //	- status: pass the exit status associated with this error.
-//	- msg: pass the additional context of the error as a string.
 //	- innner: pass the inner error to be wrapped.
```
leczb: Nit: I would just call this `err` instead of `inner`.

leczb (approved):

hermes/probe/error/error.go
evanSpendlove: Ack.

hermes/probe/error/error.go
evanSpendlove: Ack.

hermes/probe/error/error.go
```diff
@@ -0,0 +1,70 @@
...
+// Arguments:
+//	- target: pass the error to be compared with.
+// Returns:
+//	- bool: true if a match, else false.
+func (e *ProbeError) Is(target error) bool {
+	t, ok := target.(*ProbeError)
+	if !ok {
+		return false
+	}
+	return (e.Status == t.Status) && (errors.Is(e.Err, t.Err))
```
cianbgoog: This won't work if its a wrapped error containing another ProbeError which does match, will it?

evanSpendlove: If the status is the correct in the wrapped error, then it will work as `errors.Is()` unwraps and checks the inner errors.

I can add another condition: `(e.Status == t.Status || t.Status == "")` which should work for this case.

evanSpendlove: It won't indeed. I'm not sure if I'm implementing this method correctly..

hermes/probe/error/error.go
```diff
@@ -0,0 +1,70 @@
...
+	return fmt.Sprintf("%v: %v", e.Status, e.Err)
+}
+
+// Is returns true if the argument matches this object.
+// Each error that is wrapped is examined to find a match.
+// Arguments:
+//	- target: pass the error to be compared with.
+// Returns:
+//	- bool: true if a match, else false.
+func (e *ProbeError) Is(target error) bool {
```
deniadelcid: Since in a other files we use the term "target" to mean something completely different than we do here, can we pick a different variable name to not overload it? Maybe call it ''err" like we do everywhere else when naming `error` variables.

hermes/probe/error/error.go
deniadelcid: Maybe you can use [this](https://golang.org/src/syscall/syscall_unix.go?s=3011:3047#L116) as an example of how to implement it.

deniadelcid: Note that if target is `nil` (which is a valid error), this will cause a segmentation fault because neither t.Status nor t.Err will be initialized.

deniadelcid: Actually, this always triggers a segmentation fault regardless of what target is, I think.

hermes/probe/error/error.go
```diff
@@ -0,0 +1,70 @@
...
+	t, ok := target.(*ProbeError)
```
deniadelcid: Correct me if I'm wrong, but I think you intend to convert `target` into a ProbeError and assign that to `t`? However this doesn't work as you think.  It creates an empty (a.k.a. nil) `*ProbeError` and assigns that to t. This means that regardless whether `target` is nil or not,  `t` will always be nil.

Since `target` is of type `error`, I think that all you really need to do in this test is to compare `e.Err` and `t.Err`.

deniadelcid (commented):
Can you please add a test file?

-- REPLY --
The Is() function has been removed, so the only functions that remain are `New()` and `Error()`.

Would you like me to add a test file for these?

hermes/probe/error/error.go
evanSpendlove: This casts target to type `*ProbeError` and returns the newly typed object, but doesn't change the values. It doesn't create a nil error.

evanSpendlove: Example: https://play.golang.org/p/s5Jlt7BMyaF

hermes/probe/error/error.go
evanSpendlove: @leczb has pointed out that ProbeError does not need to implement an Is() method. This method is defined and implemented in the `errors` package.

evanSpendlove: I have removed this function in the latest commit.

evanSpendlove: @leczb has pointed out that ProbeError does not need to implement an Is() method. This method is defined and implemented in the errors package.

I have removed this function in the latest commit.

hermes/probe/error/error.go
evanSpendlove: @leczb has pointed out that ProbeError does not need to implement an Is() method. This method is defined and implemented in the errors package.

I have removed this function in the latest commit.

hermes/probe/error/error.go
evanSpendlove: It turns out that I don't need to implement `Is()` as it is part of the `errors` package.

hermes/probe/error/error.go
leczb: SGTM.

hermes/probe/error/error.go
```diff
@@ -30,22 +30,21 @@ import (
...
 	Err    error
 }
 
-// New returns a new ProbeError containing the error passed
-// with additional context.and the status.
+// New returns a new ProbeError containing the error and status passed.
 // Arguments:
 //	- status: pass the exit status associated with this error.
-//	- innner: pass the inner error to be wrapped.
+//	- err: pass the error to be embedded.
```
leczb: Nit: "embedding" has a particular meaning in Go. I would jus say "err: the underlying error".

leczb (approved):

cianbgoog (approved):

hermes/probe/error/error.go
```diff
@@ -0,0 +1,56 @@
...
+//
+// Error implements the ProbeError used in Hermes for recording the
+// exit status of functions.
+
+// Package error defines the ProbeError used in Hermes for recording the
+// exit status of functions.
+package error
+
+import (
+	"errors"
```
evanSpendlove: Remove unused import.

hermes/probe/error/error.go
evanSpendlove: Okay, I've changed this in the latest commit.

hermes/probe/error/error.go
evanSpendlove: Ack.

hermes/probe/probe.go
```diff
@@ -38,6 +38,33 @@ import (
...
+	Err    error
+}
+
+// NewProbeError returns a new ProbeError containing the error and status passed.
+// Arguments:
+//	- status: pass the exit status associated with this error.
+//	- err: the underlying error.
+// Returns:
+//	- ProbeError: returns a new ProbeError object containing the args passed.
+func NewProbeError(status metrics.ExitStatus, err error) *ProbeError {
```
leczb, line 54: This function doesn't do anything.
Just remove it and use &ProbeError{...} in its place.

leczb (changes requested):

cianbgoog (approved):

PR #78, opened 2020-09-21
Created fakelogger (testing purposes: read_test.go, create_test.go)

alicjakwie:


cianbgoog (commented):
Maybe I'm missing some context, but why can't we just use the actual logger?

hermes/probe/fakelogger/fakelogger.go
```diff
@@ -0,0 +1,44 @@
...
+package fakelogger
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/cloudprober/logger"
+)
+
+type FakeLogger struct {
```
leczb, line 27: This is not a fake. It's a wrapper around the real logger.
I don't understand why this is needed.

alicjakwie:
As pointed out this won't be necessary so closing this pull request!

